PR #14, opened 2016-09-27
Added an infinite loop parser and tests.

gregkatz:
This pull request adds the ability to parse `loop{}`.

A couple things: 

I had trouble with emacs and nom's syntax, so if the spacing is off, I apologize. Do you know how to make an editor handle spacing and indentation well for nom macros?

For the tests, I used tuples as statements in the body of the loop, because that's what I could get to parse.

Let me know comments and changes and I'll be glad to make them!

### DIFF
--- a/src/expr.rs
+++ b/src/expr.rs
@@ -361,7 +361,8 @@ pub mod parsing {
             // TODO: ForLoop
             // TODO: Loop
             // TODO: ForLoop
-            // TODO: Loop
+            |
+            expr_loop
             // TODO: Match
             // TODO: Closure
             |
@@ -546,6 +547,21 @@ pub mod parsing {
                 rules: BlockCheckMode::Default,
             }),
             else_block.map(Box::new),
+        ))
+    ));
+
+    named!(expr_loop -> Expr, do_parse!(
+        id: option!(
+            terminated!(
+                preceded!(
+                    punct!("'"),
+                    ident),
+                punct!(":"))) >>
+            punct!("loop") >>
+        loop_block: block >>
+        (Expr::Loop(
+            Box::new(loop_block),
+            id,
         ))
     ));
 
--- a/src/expr.rs
+++ b/src/expr.rs
@@ -337,6 +337,7 @@ pub enum BindingMode {
 pub mod parsing {
     use super::*;
     use {Ident, Ty};
+    use generics::parsing::lifetime;
     use ident::parsing::ident;
     use lit::parsing::lit;
     use nom::multispace;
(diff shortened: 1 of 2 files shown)


src/expr.rs
```diff
@@ -549,6 +550,21 @@ pub mod parsing {
...
+            terminated!(
+                preceded!(
+                    punct!("'"),
+                    ident),
+                punct!(":"))) >>
+            punct!("loop") >>
+        loop_block: block >>
+        (Expr::Loop(
+            Box::new(loop_block),
+            id,
```
dtolnay: I would like the single quote `'` to be part of the id here. That's how Rust's AST does it.

Here is what I did to check. Created `main.rs`:

```rust
fn main() {
    'label: loop {}
}
```

Ran `rustc -Z ast-json main.rs`. The following appears in the output:

```json
"node": {
    "variant": "Loop",
    "fields": [
    {
        "stmts": [],
        "id": 7,
        "rules": "Default",
        "span": {
        "lo": 29,
        "hi": 31
        }
    },
    {
        "node": "'label",
        "span": {
        "lo": 16,
        "hi": 22
        }
    }
    ]
},
```

The label is `"'label"` not just `"label"`.

src/expr.rs
```diff
@@ -549,6 +550,21 @@ pub mod parsing {
         ))
     ));
 
+    named!(expr_loop -> Expr, do_parse!(
+        id: option!(
+            terminated!(
+                preceded!(
```
dtolnay: You can simplify this in one of two ways:

1. Using [the existing lifetime parser](https://github.com/dtolnay/syn/blob/a3faf259f2620fc6cc38d81b96a6814716f119b6/src/generics.rs#L116-L121) like `terminated!(lifetime, punct!(":"))`.

    Then in the Expr::Loop at the bottom you will need to use `label.map(|lt| lt.ident)` to pull out the ident.

    You will need to add `use generics::parsing::lifetime` [up here](https://github.com/dtolnay/syn/blob/a3faf259f2620fc6cc38d81b96a6814716f119b6/src/expr.rs#L340-L343) to be able to use it.

2. Using the `delimited` helper which is a combination of `preceded` and `terminated`:

    `delimited!(punct("'"), ident, punct(":"))`.

    Note that in this case you need to do `format!("'{}", id).into()` below in order to address my other comment.

Up to you which one you find nicer.

src/expr.rs
```diff
@@ -549,6 +550,21 @@ pub mod parsing {
...
+                    punct!("'"),
+                    ident),
+                punct!(":"))) >>
```
dtolnay: This should be unindented to be at the same level as "id" above and "loop_body" below.

src/expr.rs
```diff
@@ -549,6 +550,21 @@ pub mod parsing {
...
+            punct!("loop") >>
```
dtolnay: Same indentation here as the previous line.

src/expr.rs
dtolnay: Never mind this comment, I misread the parentheses. The one on the next line applies though.

src/expr.rs
dtolnay: Same indentation as the *next* line.

gregkatz:
Thanks for the comments! I added another commit to hopefully address them. I had never thought of a loop label as a lifetime, but it makes sense when you think about it.

PR #15, opened 2016-09-28
Added while loop parser and (currently ignored) tests.

gregkatz:
Hi David,

This is almost identical to the infinite loop parser. I wrote tests but I left them disabled until a little more of the language can be parsed. (I couldn't come up with a test that makes sense with what is currently available.)

If I get time to do some more of these, would you prefer them in a larger pull request? Maybe two or three at a time? Or do you prefer to keep your PRs small?

Let me know your comments and corrections! Thanks for taking the time for review!

--Greg

### DIFF
--- a/src/expr.rs
+++ b/src/expr.rs
@@ -357,7 +357,8 @@ pub mod parsing {
             |
             expr_if
             // TODO: IfLet
-            // TODO: While
+            |
+            expr_while
             // TODO: WhileLet
             // TODO: ForLoop
             // TODO: Loop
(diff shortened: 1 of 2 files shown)


dtolnay:
Perfect! Keep them coming :heart_eyes: 

I don't have a preference between smaller or larger PRs so whatever is easiest for you. The only thing is I think having multiple of your PRs open at the same time will be confusing because we will end up with conflicts between them. So if you open a PR and I don't get to it until several hours later, just keep pushing your work as additional commits in the same PR.

Issue #42, opened 2016-10-17
Add `From` conversions for syn::Lit


dtolnay:
For example:

```rust
let s: String = /* ... */;
let lit: Lit = s.into();
```

There should be conversions for all the Lit variants.

PR #50, opened 2016-10-19
Add a bunch of From impls for Lit

killercup:
I had a bit of time earlier and wanted to dig into syn's code base. And why not implement a small thing at the same time? :)

Is this approximately what you meant in #42?

I haven't added any new tests for this (should I?), but used some of the impls in the code.


dtolnay:
Exactly what I had in mind, thanks Pascal.

PR #52, opened 2016-10-20
Add parse_token_trees.

SimonSapin:
Token trees are much simpler than an AST. This makes them easier to work with when they are sufficient for the task at hand, such as for example [expanding a procedural macro](https://github.com/servo/html5ever/pull/217).


dtolnay:
Thanks! I released this as syn 0.9.2.

Issue #5, opened 2016-09-05
Add quasi-quote macros that return AST objects


dtolnay:
Delegate to the `quote!` macro from the [`quote` crate](https://github.com/dtolnay/quote), then parse the resulting tokens into the appropriate AST type. Example usage:

``` rust
let attr: Attribute = quote_attribute!{ #[derive(Debug, Clone, #a, #b)] };
```

SimonSapin:
Rather than a large number of macros for each type, maybe a method available on each type would be a nicer API?

```rust
let attr = Attribute::parse(quote!(#[derive(Debug, Clone, #a, #b)])).unwrap();
```

These methods could also replace the ad-hoc top-level `parse_*` functions, if they accept a generic parameter with some trait implemented by both `quote::Tokens` and `&str` (and maybe `&[syn::TonkenTree]` / `Vec<TokenTree>` / `TokenTree`? For dealing with expression/items macros.)

dtolnay:
I like it but I have two concerns:

- I think `Attribute::parse` puts too much emphasis on the fact that parsing is happening, which is not what I want the user and readers of the user's code to focus on in these cases. Everyone knows parsing is something that can fail, so you need to unwrap or `try!(Attribute::parse(...))`. But if the thing you are parsing is right there inside a `quote!(...)` macro, you should write it in a way that you know will parse i.e. anything that doesn't parse is a bug in your program, not an expected failure that you can handle at runtime. I think `quote_attribute!(...)` captures this distinction better.
- Once procedural macros are stable, the `quote_attribute!` approach allows us to do the parsing at compile time rather than run time in most cases. As I mentioned in another ticket I am not worried about performance here so that is not why I am excited about this. Instead, parsing at compile time would let us validate the thing you are quoting, and fail to compile your program if it looks wrong. We can't ever do this with the `Attribute::parse(quote!(...))` approach.

I am not settled on the `quote_attribute!` approach yet but these are some of the advantages.

I do like `Attribute::parse` as an alternative to the current style of `syn::parse_attribute`, will think about the tradeoffs there.

SimonSapin:
Could `quote_attribute!` do compile-time parsing while keeping the generality of "Anything that implements `ToTokens` can be used with `#foo` interpolation" ?

Anyway, just to write down the idea, the traits I mentioned before could look like this:

```rust
pub trait Parse: Sized {
    fn parse<I: ToSourceString>(input: I) -> Result<Self, String>;
}

[...]

PR #62, opened 2016-10-31
Path constructors and WhereClause::push_simple_bound

SimonSapin:


SimonSapin:
These would simplify https://github.com/servo/heapsize/pull/64#discussion_r85831245. But maybe `add_simple_bound` is too specific?

dtolnay:
Are there reasons other than performance that you would prefer this over the approach using https://github.com/dtolnay/syn/issues/5?

```rust
where_clause.push_simple_bound(t, Path::new_global(&["heapsize", "HeapSizeOf"]));

where_clause.predicates.push(quote_where_predicate!(#t: ::heapsize::HeapSizeOf));
```

The other one is more flexible in that it supports paths with type parameters (#t: Into\<u8>) and lifetime bounds (#t: 'a) and multiple bounds (#t: A + B + #c) using exactly the same syntax.

Performance is a valid reason but not really a priority right now.

SimonSapin:
I’m not too worried about performance (though I don’t have any number on time spent building and running compile-time stuff, maybe it adds up).

This PR was my immediate idea to try to simplify the code I wrote for `derive(HeapSizeOf)` but yeah, maybe a more general approach like #5 is preferable.

dtolnay:
Let's go with the quasi-quoting approach.

PR #64, opened 2016-11-09
re-export ExprKind

KodrAus:
Super simple one, just re-exporting `ExprKind` in the root. Or is there a specific reason this wasn't included?


dtolnay:
Thanks! No reason, I messed up when I introduced ExprKind. I released this as 0.10.1.

PR #79, opened 2017-01-22
Add support for the 'full' feature to 'visit'

mystor:
I'm not sure if it's the cleanest code I've ever written, but it seems to work well enough. 

Not sure what the best solution for `ConstExpr::Other` and `Ty::Mac` is, the current workaround I threw together for this patch isn't super nice.


dtolnay:
Thanks! I am so glad someone else needed this before I did :smile_cat:. I will get a release ready tomorrow for you to depend on.

dtolnay:
I released this change in [0.11.0](https://github.com/dtolnay/syn/releases/tag/0.11.0).

PR #82, opened 2017-01-24
Expose the nom parsers in a public `parsing::` submodule

### LINKED ISSUE #81: Publicly expose parsing functions which don't consume the entire string.
Currently if I am trying to parse, say, a macro body with custom syntax like:

```rust
foo!(a as "a", b as "b")
```

So, I have the input string `'a as "a", b as "b"'` and I want to parse it, I call `syn::parse_ident(s)` to parse the first ident from the string, and get an error, as there are characters which follow that ident. It would be nice to also have access to methods (say `syn::partial_parse_ident(&str) -> Result<(Ident, &str), String>`) which return the remaining string slice in addition to the ident which was parsed.

This could also be done by changing the signature of `syn::parse_ident` to something more like:

```rust
enum ParseError<'a, T> {
    Partial(&'a str, T),
    Failed(String)
}
impl<'a, T> Error for ParseError<'a, T> {..}

fn parse_ident<'a>(input: &'a str) -> Result<Ident, ParseError<'a, Ident>>
```

Which would allow me to extract the parsed value and remaining string data from the error value.

Implementing this shouldn't be difficult, but I would like @dtolnay's input on the design before I do so.

mystor:
I quickly threw this together, because this solution is super easy and straightforward.

Fixes #81.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -9,11 +9,7 @@ extern crate unicode_xid;
 
 #[cfg(feature = "parsing")]
 #[macro_use]
-mod nom;
-
-#[cfg(feature = "parsing")]
-#[macro_use]
-mod helper;
+extern crate syn_nom as nom;
 
 #[cfg(feature = "aster")]
 pub mod aster;
@@ -73,9 +69,6 @@ pub type MacroInput = DeriveInput;
 mod op;
 pub use op::{BinOp, UnOp};
 
-#[cfg(feature = "parsing")]
-mod space;
-
 mod ty;
 pub use ty::{Abi, AngleBracketedParameterData, BareFnArg, BareFnTy, FunctionRetTy, MutTy,
              Mutability, ParenthesizedParameterData, Path, PathParameters, PathSegment,
@@ -90,8 +83,8 @@ pub use parsing::*;
 #[cfg(feature = "parsing")]
 mod parsing {
     use super::*;
-    use {derive, generics, ident, mac, space, ty};
-    use nom::IResult;
+    use {derive, generics, ident, mac, ty};
+    use nom::{space, IResult};
 
     #[cfg(feature = "full")]
     use {expr, item, krate};
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -163,3 +163,53 @@ mod parsing {
         }
     }
 }
+
+#[cfg(feature = "parsing")]
+pub mod parser {
+    //! This module contains a set of exported nom parsers which can be used to
+    //! build nom parsers for custom grammars when used alongside the `syn_nom`
+    //! crate.
+    //!
+    //! `syn` uses its own custom fork of `nom`, `syn_nom`, which is smaller, and
+    //! thus improves build speeds. This should be used instead of `nom` when
+    //! building parsers using these parsers.
+
+    #[cfg(feature = "full")]
+    pub use krate::parsing::krate;
+
+    #[cfg(feature = "full")]
+    pub use item::parsing::item;
+
+    #[cfg(feature = "full")]
+    pub use item::parsing::items;
+
+    #[cfg(feature = "full")]
+    pub use expr::parsing::expr;
+
+    pub use lit::parsing::lit;
+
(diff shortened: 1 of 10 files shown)


src/lib.rs
```diff
@@ -167,3 +167,55 @@ mod parsing {
...
+
+    pub use lit::parsing::byte as byte_lit;
+
+    pub use lit::parsing::character as char_lit;
+
+    pub use lit::parsing::float as float_lit;
+
+    pub use lit::parsing::int as int_lit;
+
+    pub use lit::parsing::boolean as bool_lit;
```
dtolnay: I would prefer to have these lit parsers return their own individual types - it seems silly to parse a bool_lit and then have to pattern-match on a Lit to read the bool out of it. There should be a parser called "boolean" that gives you a `bool`.

For string / byte string / int / float we will need some sort of wrapper struct like `struct StrLit(String, StrStyle)`.

src/lib.rs
```diff
@@ -167,3 +167,55 @@ mod parsing {
...
+    //! crate.
+    //!
+    //! `syn` uses its own custom fork of `nom`, `syn_nom`, which is smaller, and
+    //! thus improves build speeds. This should be used instead of `nom` when
+    //! building parsers using these parsers.
+
+    pub use macro_input::parsing::macro_input;
+
+    #[cfg(feature = "full")]
+    pub use krate::parsing::krate;
```
dtolnay: I don't think this is one that people will need in their procedural macros, let's leave it out.

src/lib.rs
```diff
@@ -167,3 +167,55 @@ mod parsing {
...
+pub mod parsing {
+    //! This module contains a set of exported nom parsers which can be used to
+    //! build nom parsers for custom grammars when used alongside the `syn_nom`
+    //! crate.
+    //!
+    //! `syn` uses its own custom fork of `nom`, `syn_nom`, which is smaller, and
+    //! thus improves build speeds. This should be used instead of `nom` when
+    //! building parsers using these parsers.
+
+    pub use macro_input::parsing::macro_input;
```
dtolnay: Let's leave this out, this one is specific to the macros 1.1 custom derive use case and isn't something people will need for other procedural macros.

syn_nom/src/lib.rs
```diff
@@ -10,26 +10,29 @@ pub enum IResult<I, O> {
     Error,
 }
```
dtolnay: Please also move all the macros in src/helper.rs - you won't get very far without those, especially `punct!` and `keyword!`. Keep them in a separate file though to track that they are not from upstream nom.

src/lib.rs
mystor: I haven't done this change yet

utkarshkukreti:
@dtolnay do you plan to release a new version of syn with these changes soon now that #87 has been fixed? Thanks!

dtolnay:
@utkarshkukreti what do you need it for? I was not in a hurry to release because I didn't think this would be useful until function-like procedural macros land in nightly.

utkarshkukreti:
@dtolnay I'm writing a template engine and need to be able to parse the longest prefix of a str which is a valid expression. I was using a pretty inefficient way to solve this earlier -- trying to parse an expression at every whitespace and taking the last valid one (which was very slow) before I found this pull request.

Thanks for the release!

One quick question: is there any reason you haven't exposed [the `-> Stmt` parsers](https://github.com/dtolnay/syn/blob/2ce94ed368498a0587f71ea0ada1f8fb91ece08e/src/expr.rs#L845)? Would you accept a PR to make them public? I wouldn't need a release right away -- I'm still at least a week away from doing a public release of my crate.

dtolnay:
I agree, perhaps we started with too conservative of a choice of functions. I added `stmt` as well as `pat` and `block` in 1bf19134bd136894d7b0ae1195f629e28e6827e2. Let me know if there are any others that would be useful to you.

utkarshkukreti:
Thanks, I definitely needed `pat` as well. I will let you know if I need more!

PR #83, opened 2017-01-25
Update unicode-xid dependency.

SimonSapin:
This helps Servo avoid having two copies of it.


SimonSapin:
Oh, except a lot of our dependencies are still using syn 0.10.x… would you be ok with back-porting this change and publishing 0.10.7?

I’ve changed the PR so that it starts at 0.10.6, and added a merge commit with current master to resolve the conflict. If this works for you, you can checkout the intermediate commit and run `cargo publish` from there.

dtolnay:
Thanks, I released 0.10.7 and 0.11.1 with this change.

PR #85, opened 2017-01-26
Implement Folder

### LINKED ISSUE #43: Implement `fold`
Here is how libsyntax does it: https://github.com/rust-lang/rust/blob/master/src/libsyntax/fold.rs

cc @SimonSapin

gnzlbg:
Closes #43 .

Adds a new feature: "fold". 

The `Folder` trait allows implementing AST->AST transformations.

If the way it is implemented looks good, adding the documentation is trivial, but adding some tests will take some work.

### DIFF
--- a/src/fold.rs
+++ b/src/fold.rs
@@ -1,0 +1,943 @@
+// Adapted from libsyntax.
+
+//! A Folder represents an AST->AST fold; it accepts an AST piece,
+//! and returns a piece of the same type. 
+
+use super::*;
+
+/// AST->AST fold.
+///
+/// Each method of the Folder trait is a hook to be potentially overridden. Each
+/// method's default implementation recursively visits the substructure of the
+/// input via the `noop_fold` methods, which perform an "identity fold", that
+/// is, they return the same structure that they are given (for example the
+/// `fold_crate` method by default calls `fold::noop_fold_crate`).
+///
+/// If you want to ensure that your code handles every variant explicitly, you
+/// need to override each method and monitor future changes to `Folder` in case
+/// a new method with a new default implementation gets introduced.
+pub trait Folder: Sized {
+    // Any additions to this trait should happen in form
+    // of a call to a public `noop_*` function that only calls
+    // out to the folder again, not other `noop_*` functions.
+    //
+    // This is a necessary API workaround to the problem of not
+    // being able to call out to the super default method
+    // in an overridden default method.
+
+    fn fold_ident(&mut self, _ident: Ident) -> Ident {
+        noop_fold_ident(self, _ident)
+    }
+    fn fold_derive_input(&mut self, derive_input: DeriveInput) -> DeriveInput {
+        noop_fold_derive_input(self, derive_input)
+    }
+    fn fold_ty(&mut self, ty: Ty) -> Ty {
+        noop_fold_ty(self, ty)
+    }
+    fn fold_generics(&mut self, generics: Generics) -> Generics {
+        noop_fold_generics(self, generics)
+    }
+    fn fold_ty_param_bound(&mut self, bound: TyParamBound) -> TyParamBound {
+        noop_fold_ty_param_bound(self, bound)
+    }
+    fn fold_poly_trait_ref(&mut self, trait_ref: PolyTraitRef) -> PolyTraitRef {
+        noop_fold_poly_trait_ref(self, trait_ref)
+    }
+    fn fold_variant_data(&mut self, data: VariantData) -> VariantData {
+        noop_fold_variant_data(self, data)
+    }
+    fn fold_field(&mut self, field: Field) -> Field {
+        noop_fold_field(self, field)
+    }
+    fn fold_variant(&mut self, variant: Variant) -> Variant {
+        noop_fold_variant(self, variant)
+    }
+    fn fold_lifetime(&mut self, _lifetime: Lifetime) -> Lifetime {
+        noop_fold_lifetime(self, _lifetime)
+    }
+    fn fold_lifetime_def(&mut self, lifetime: LifetimeDef) -> LifetimeDef {
+        noop_fold_lifetime_def(self, lifetime)
+    }
+    fn fold_path(&mut self, path: Path) -> Path {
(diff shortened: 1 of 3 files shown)


dtolnay:
Thanks! This is exactly what I wanted. I don't have time to review today but I will get to it tomorrow.

gnzlbg:
@dtolnay I think we should strive to provide a `MacroExpander` `Folder` in `syn` that is able to expand all macros (otherwise `Folder` is really hard to use correctly for anything). I guess that basically would require reimplementing [`libsyntax::ext::expand`](https://github.com/rust-lang/rust/blob/aedb49cbc9977fee56bb51a20980b380aca53659/src/libsyntax/ext/expand.rs) on top of `Folder`.

mystor:
@gnzlbg I think that a `MacroExpander` `Folder` would fit more in an external helper crate which can be used with `syn` than directly inside of syn. I feel like `syn` should keep itself mostly concerned only with parsing, and not with how the results of those parses will be used.

The main reason why I think `fold` and `visit` are in `syn` is because their implementation is so tightly tied to each version of `syn`'s internal representation.

This is all up to @dtolnay though - so if he thinks that `MacroExpander` has a place in `syn`, then so be it :).

gnzlbg:
@mystor I haven't written `MacroExpander` yet, so I don't know how tightly it would need to be integrated with `syn`. Ideally one could write it as a `Folder`, but this is not the way `libsyntax` does it. 

So... I am a bit stumbled about how to proceed. I don't know enough about macro expansion to know if even attempting to write it as a `Folder` make sense. Maybe it should be "it's own thing" like in libsyntax, and more tightly integrated with parsing, since macro expansion should just be a `TokenTree -> TokenTree` transformation, instead of an `AST -> AST` transformation.

dtolnay:
I don't think `syn` needs to be in the business of pretending that we can fully expand macros in your crate. The reality is that only rustc can make that guarantee. Procedural macros need to accept this and work in a way that handles unexpanded macros correctly.

I would be in favor of having `fold_mac` implemented as `noop_fold_mac` by default. The situation in libsyntax is a bit different - if rustc is folding something that contains unexpanded macros, that is probably a bug. In `syn` the expected use cases are ones where macro expansion is not available so folders need to be written in a way that supports unexpanded macros. In other words finding unexpanded macros in the input is the expected state of the world and does not indicate a bug the way it would in rustc.

src/fold.rs
```diff
@@ -0,0 +1,984 @@
...
+/// Each method of the Folder trait is a hook to be potentially overridden. Each
+/// method's default implementation recursively visits the substructure of the
+/// input via the `noop_fold` methods, which perform an "identity fold", that
+/// is, they return the same structure that they are given (for example the
+/// `fold_crate` method by default calls `fold::noop_fold_crate`).
+///
+/// If you want to ensure that your code handles every variant explicitly, you
+/// need to override each method and monitor future changes to `Folder` in case
+/// a new method with a new default implementation gets introduced.
+pub trait Folder: Sized {
```
dtolnay: We don't need this `Sized` bound. Unclear why libsyntax has it.

src/fold.rs
```diff
@@ -0,0 +1,984 @@
...
+    fn fold_ty(&mut self, ty: Ty) -> Ty {
+        noop_fold_ty(self, ty)
+    }
+    fn fold_generics(&mut self, generics: Generics) -> Generics {
+        noop_fold_generics(self, generics)
+    }
+    fn fold_ty_param_bound(&mut self, bound: TyParamBound) -> TyParamBound {
+        noop_fold_ty_param_bound(self, bound)
+    }
+    fn fold_poly_trait_ref(&mut self,
```
dtolnay: Let's take only the PolyTraitRef as input here, the way libsyntax does.

src/fold.rs
```diff
@@ -0,0 +1,984 @@
...
+                           trait_ref: PolyTraitRef,
+                           modifier: TraitBoundModifier)
+                           -> PolyTraitRef {
+        noop_fold_poly_trait_ref(self, trait_ref, modifier)
+    }
+    fn fold_variant_data(&mut self,
```
dtolnay: Let's take only the VariantData as input.

src/fold.rs
```diff
@@ -0,0 +1,984 @@
...
+                         data: VariantData,
+                         _ident: Ident,
+                         _generics: Generics)
+                         -> VariantData {
+        noop_fold_variant_data(self, data, _ident, _generics)
+    }
+    fn fold_field(&mut self, field: Field) -> Field {
+        noop_fold_field(self, field)
+    }
+    fn fold_variant(&mut self, variant: Variant, generics: Generics) -> Variant {
```
dtolnay: Let's take only the Variant as input.

src/fold.rs
```diff
@@ -0,0 +1,984 @@
...
+    #[cfg(feature = "full")]
+    fn fold_method_sig(&mut self, method_sig: MethodSig) -> MethodSig {
+        noop_fold_method_sig(self, method_sig)
+    }
+    #[cfg(feature = "full")]
+    fn fold_stmt(&mut self, stmt: Stmt) -> Stmt {
+        noop_fold_stmt(self, stmt)
+    }
+
+    /// Prefer overriding fold_stmt instead!
```
dtolnay: Why? `Block` shows up in lots of other places.

src/fold.rs
gnzlbg: `dyn::Visitor` has this as well

src/fold.rs
gnzlbg: I think one often wants to fold over a `Block`, so `Folder` should offer a `.fold_block` method, but since a `Block` is just a vector of `Stmt`s, one typically wants to override `fold_stmt` instead (and not `fold_block`).

gnzlbg:
@dtolnay I've fixed all issues you mentioned :)

src/fold.rs
dtolnay: I think this is just a matter of accepting `<F: ?Sized + Folder>` in the noop functions. I will see if I can make it work.

Issue #87, opened 2017-01-28
Document the parsing macros


dtolnay:
Before including this in a release, I would like to document all of the parsers with realistic procedural-macro-appropriate examples. @mystor would you be able to help me work through these?

- [x] alt
- [x] call
- [x] cond
- [x] cond_reduce
- [x] delimited
- [x] do_parse
- [x] epsilon
- [x] keyword
- [x] many0
- [x] map
- [x] named
- [x] not
- [x] opt_vec
- [x] option
- [x] peek
- [x] preceded
- [x] punct
- [x] separated_list
- [x] separated_nonempty_list
- [x] switch
- [x] tag
- [x] take_until
- [x] take_while1
- [x] terminated
- [x] terminated_list
- [x] tuple
- [x] value

I started you off with two examples of what I have in mind but feel free to invent your own approach if you want.

> One or more of something separated by some separator.
>
> - **Syntax:** `separated_nonempty_list!(SEPARATOR, THING)`
> - **Output:** `Vec<THING>`
>
> ```rust
> extern crate syn;
> #[macro_use] extern crate synom;
>
> use syn::Ty;
> use syn::parse::ty;
>
> // One or more Rust types separated by commas.
> named!(comma_separated_types -> Vec<Ty>,
>     separated_nonempty_list!(
>         punct!(","),
>         ty));
>
> fn main() {
>     let input = "&str, Map<K, V>, String";
>
>     let parsed = comma_separated_types(input).expect("comma-separated types");
>
>     assert_eq!(parsed.len(), 3);
>     println!("{:?}", parsed);
> }
> ```

---

> Value surrounded by a pair of delimiters.
>
> - **Syntax:** `delimited!(OPEN, THING, CLOSE)`
> - **Output:** `THING`
>
> ```rust
> extern crate syn;
> #[macro_use] extern crate synom;
>
> use syn::Expr;
> use syn::parse::expr;
>
> // An expression surrounded by [[ ... ]].
> named!(double_bracket_expr -> Expr,
>     delimited!(
>         punct!("[["),
>         expr,
>         punct!("]]")));
>
> fn main() {
>     let input = "[[ 1 + 1 ]]";
>
>     let parsed = double_bracket_expr(input).expect("double bracket expr");
>
>     println!("{:?}", parsed);
> }
> ```

dtolnay:
Fixed in #91 and #93.

PR #91, opened 2017-02-05
Document the synom parser macros

mystor:
This documents the rest of the parser macros, fixing #87.

I did it all in one go, and I didn't read over it, so there may be some mistakes. Please point them out when you look over it.

### DIFF
--- a/synom/src/helper.rs
+++ b/synom/src/helper.rs
@@ -1,6 +1,22 @@
 use IResult;
 use space::{skip_whitespace, word_break};
 
+/// Parse a piece of punctuation, skipping whitespace before it.
+///
+/// - **Syntax:** `punct!("...")`
+/// - **Output:** `&str`
+///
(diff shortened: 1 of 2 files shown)


dtolnay:
Awesome! I made it partway through and I will continue tomorrow.

One thing I realized i need to add myself is an overview of the whitespace-handling strategy. I think the way we do it is far better (for our use case) than nom's `ws` macro. In short: we have a small number of building-block parsers (punct, keyword, lit, doc comment) that all ignore leading whitespace. As long as every other parser is combinated out of only these building blocks, nothing else needs to worry about whitespace anywhere else. And it is great for compile time - the cost in compile time that we pay for correct whitespace handling is basically four function calls, one in each building-block parser, rather than pervasive `match` expressions checking for whitespace between every possible pair of tokens like `ws` would do.

PR #92, opened 2017-02-07
Add Spans

### LINKED ISSUE #41: Implement spans
This is a requirement for implementing something like rustfmt against syn.

mystor:
This is more of a WIP. I'm hoping that @dtolnay can give me feedback on the approach.

This is built on top of #91, so it has that commit included.

Fixes #41

### DIFF
--- a/src/attr.rs
+++ b/src/attr.rs
@@ -126,9 +126,12 @@ pub mod parsing {
                 value: MetaItem::NameValue(
                     "doc".into(),
                     Lit {
-                        span: content.1,
-                        ..format!("//!{}", content.0).into()
-                    },
+                        span: Span {
+                            lo: content.span.lo - 3, // include the '//!'
+                            hi: content.span.hi,
+                        },
+                        .. format!("//!{}", content.node).into()
+                    }
                 ),
                 is_sugared_doc: true,
                 span: DUMMY_SPAN,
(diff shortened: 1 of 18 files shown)


dtolnay:
I think this is a great approach! Amazing work.

One minor thing is I would prefer to see less `.0` and `.1`. I think this is the culprit:

```rust
::synom::IResult::Done(rest, (val, span))
```

How would you feel about a struct with `node` and `span` [to match rustc?](https://github.com/rust-lang/rust/blob/1.15.0/src/libsyntax/codemap.rs#L59-L62)

I am traveling for job interviews and have been stressed / overbooked, should be back to normal productivity by Wednesday.

mystor:
I added a Spanned<T> generic type earlier, and just had an idea for how to reduce the breaking-changeness of this chance, to potentially allow us to land it sooner / more easily.

The idea would be to have the `span` module have 2 implementations. The current one would be used when the `span` feature is enabled, and a dummy implementation would be used when it isn't:

```rust
pub type Spanned<T> = T;

macro_rules! spanned {
    ($i:expr, $submac:ident!( $($args:tt)* )) => {
        $submac!($i, $($args)*)
    }
}
```

Instead of adding a `span` field to different type definitions, we pass around `Spanned<Ty>` `Spanned<Ident>` etc. which will act exactly the same as the base type in the current implementation, but will have spans attached to them when the `span` feature is enabled.

This way we can add support for spans when people want them, without breaking code for consumers who are currently using syn with custom derive, where spans aren't super useful right now.

What do you think about that idea @dtolnay?

dtolnay:
The thing with cfgs is they are required to be purely additive. If I have code in my crate that looks like:

```rust
let e = Expr { node: kind, attrs: vec![] };
```

and I add a dependency on a crate that uses syn with spans enabled, Cargo will build my crate with spans enabled and it won't work.

dtolnay:
This is happening for real, using the compiler's spans! #142

PR #98, opened 2017-02-27
Parse ConstExpr::Other

SimonSapin:


src/constant.rs
```diff
@@ -94,6 +98,14 @@ pub mod parsing {
 
     named!(expr_path -> ConstExpr, map!(path, ConstExpr::Path));
 
+    #[cfg(feature = "full")]
+    named!(expr_other -> ConstExpr, map!(expr, ConstExpr::Other));
+
+    #[cfg(not(feature = "full"))]
+    fn expr_other(_: &str) -> IResult<&str, ConstExpr> {
+        IResult::Error
```
SimonSapin: Is there a macro to do this?

PR #102, opened 2017-03-01
fix missing '$' before 'separgs' in expansion of separated_list

mindsbackyard:
The macro expansion of `separated_list` is broken as a '$' sign is missing.
The error is not visible from the CI builds. Probably because the macro is never expanded in the tests.
An example triggering the error is:
```rust
named!(comma_separated_types -> Vec<syn::Ty>,
    separated_list!(punct!(","), syn::parse::ty)
);
```

### DIFF
--- a/synom/src/helper.rs
+++ b/synom/src/helper.rs
@@ -368,7 +368,7 @@ macro_rules! separated_list {
     }};
 
     ($i:expr, $sepmac:ident!( $($separgs:tt)* ), $f:expr) => {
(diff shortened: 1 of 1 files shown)


dtolnay:
Good catch. I fixed three others in 6e825ca89225827d3a3b9c55a89d2a2dede8851d.

dtolnay:
I published a release with your fix.

mindsbackyard:
Wow, pretty fast. Thanks for the fixes!

PR #104, opened 2017-03-05
Fixes #103; more impls for Path.

### LINKED ISSUE #103: FromIterator<Into<PathSegment>>, Ord for Path
Willing to write this up at some point, but figured I should open an issue for it.

clarcharr:


clarcharr:
Note to self: will have to implement `Ord` for lots of other structs to make this work.

dtolnay:
Can you give some motivation for why you need each of these impls? How are you planning on using this?

clarcharr:
I'm working on a tool to parse out impls from crates and I was going to use the `FromIterator` implementation to make returning module paths a lot simpler; an iterator of idents will become a path.

As far as the `Ord` implementations go, I'd like to ultimately sort the impls I end up with and would rather not have to cast all the idents as strings in order to sort them.

dtolnay:
Why is !global an obvious choice for the FromIterator impl? If it is obvious only in the context of your project, it may fit better as a helper function in your crate.

clarcharr:
It seems obvious to me when you consider `global` to be an empty path segment:

`["", "std"] < ["", "std", "io"] < ["std"] < ["std", "io"]`

I personally think that it's beneficial to have an absolute sorting convention for paths, even if this convention isn't the most intuitive.

Another alternative could be to sort them last (i.e. derive `Ord`) because if I'm sorting an impl list, it makes most sense to put `impl ... for u8` before `impl ... for ::serde::ser::Serialize`.

I do think that a standard ordering, even if it's not the most intuitive, would be useful to have in general. I'm fine with implementing it in my own crate if that's not desired, however.

dtolnay:
Oh for Ord I agree that whether global comes first or last doesn't matter so much. I was asking about the choice of having FromIterator build !global paths.

One thing I don't like here is that paths don't seem to me like they have an ordering conceptually. For example, which one of these paths comes first and why?

```rust
std::option::Option<Self::Item>
std::option::Option<&Self::Item>
std::option::Option<<Self as Iterator>::Item>
```

If the answer is that they happened to be represented in a particular order in syn's AST, I would prefer not to add Ord impls based on that alone.

If what you need in your crate is really just an arbitrary order, you can define a significantly more intuitive one in your own crate like this:

```rust
#[macro_use]
extern crate quote;
extern crate syn;

use std::cmp::Ordering;
use std::collections::BTreeSet as Set;
use syn::Path;

#[derive(PartialEq, Eq, Debug)]
struct LexicographicPath(syn::Path);

impl PartialOrd for LexicographicPath {
    fn partial_cmp(&self, other: &LexicographicPath) -> Option<Ordering> {
        Some(self.cmp(other))
    }
}

impl Ord for LexicographicPath {
    fn cmp(&self, other: &LexicographicPath) -> Ordering {
        let a = &self.0;
        let b = &other.0;
        quote!(#a).as_str().cmp(quote!(#b).as_str())
    }
}

fn main() {
    let mut set = Set::new();
    set.insert(LexicographicPath(
        syn::parse_path("std::option::Option<Self::Item>").unwrap()));
    set.insert(LexicographicPath(
        syn::parse_path("std::option::Option<&Self::Item>").unwrap()));
    set.insert(LexicographicPath(
        syn::parse_path("std::option::Option<<Self as Iterator>::Item>").unwrap()));
    println!("{:?}", set);
}
```

clarcharr:
I think I'm going to close this PR for now and open something else up if I come up with a better alternative.

PR #105, opened 2017-03-06
public attribute parsers

colin-kiegel:
I would like to have access to attribute parsing.

My use case is this: I am writing unit tests for `derive_builder`.

```rust
#[test]
fn setter_enabled() {
    let attr = syn::Attribute{
        style: syn::AttrStyle::Outer,
        value: syn::MetaItem::Word(syn::Ident::new("foo")),
        is_sugared_doc: false,
    };

    let field = BuilderField {
        attrs: &vec![&attr],
        /* ... */
    };

    assert_eq!(quote!(#field), quote!(
        #[foo] pub foo: ::std::option::Option<String>,
    ));
}
```

would simplify to

```rust
#[test]
fn setter_enabled() {
    let attr = syn::parse_outer_attr("#[foo]").unwrap();

    let field = BuilderField {
        attrs: &vec![&attr],
        /* ... */
    };

    assert_eq!(quote!(#field), quote!(
        #[foo] pub foo: ::std::option::Option<String>,
    ));
}
```

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -87,7 +87,7 @@ pub use parsing::*;
 #[cfg(feature = "parsing")]
 mod parsing {
     use super::*;
-    use {derive, generics, ident, mac, ty};
+    use {derive, generics, ident, mac, ty, attr};
     use synom::{space, IResult};
 
     #[cfg(feature = "full")]
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -141,6 +141,15 @@ mod parsing {
         unwrap("type parameter bound", generics::parsing::ty_param_bound, input)
     }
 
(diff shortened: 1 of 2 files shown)


clarcharr:
Why not just use quote?

colin-kiegel:
@clarcharr I _am_ using `quote`. Do you wonder why I use the `BuilderField` struct?

One of the reasons is that I can write unit tests this way. The `derive_builder` crate already has a lot of options (and it's getting more). `BuilderField` is a struct that implements `ToTokens` and uses `quote` internally. But from the outside it provides a clean API which is easy to test.

dtolnay:
Thanks, I published these in 0.11.9.

PR #106, opened 2017-03-10
Add some doc comments.

clarcharr:
Didn't get everything, but I added some. Also added some newlines between fields/variants for readability.


dtolnay:
Thanks! These AST types are mostly lifted from [`libsyntax/ast.rs`](https://github.com/rust-lang/rust/blob/master/src/libsyntax/ast.rs). I'm sure they would appreciate the equivalent changes over there.

clarcharr:
Will do!

PR #107, opened 2017-03-12
Replace time with std::time

serprex:


serprex:
https://github.com/rust-lang/cargo/issues/3819

PR #112, opened 2017-04-20
Added From<Ident> for TyParam

TedDriggs:
Small convenience method for creating a `TyParam`.


dtolnay:
Thanks! Makes sense to me.

PR #122, opened 2017-04-21
tests/rust submodule update + test_round_trip fixes that it depends on

Arnavion:


dtolnay:
Wow thank you so much for taking care of this!

PR #127, opened 2017-04-25
Support parsing catch expressions.

### LINKED ISSUE #109: Catch expressions added to AST
These should be added in syn too.

Arnavion:
Fixes #109 

---

Before this change, `tests/rust/src/test/run-pass/catch-expr.rs` needed to be ignored because `syn` could not parse it, so it would fail the round-trip test. `syn` *can* parse it now with this change, but `syntex` still can't so the round-trip can't be tested anyway. So I've stopped ignoring the file, but also added its contents as a new test that validates the relevant parts of the parsed AST.

### DIFF
--- a/src/expr.rs
+++ b/src/expr.rs
@@ -176,6 +176,11 @@ pub enum ExprKind {
 
     /// `expr?`
     Try(Box<Expr>),
+
+    /// A catch expression.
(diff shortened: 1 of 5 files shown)


dtolnay:
Thank you!

Issue #130, opened 2017-05-01
Use #[deprecated] attribute on parse_macro_input


dtolnay:

dtolnay:
Fixed in #131.

PR #131, opened 2017-05-01
Added FromStr implementations to mirror the parse_* functions

TedDriggs:
Also added #[deprecated] to parse_macro_input to pop warnings for those still using it (see #130).


dtolnay:
Thanks!

TedDriggs:
@dtolnay if there isn't an ETA for 0.12, can this, #139, and #134 ship in an update to 0.11?

dtolnay:
#137 and #142 are going to involve reworking how all of this is exposed, so I would prefer not to push any of this into 0.11.

PR #134, opened 2017-05-16
Add method from 0.12 to syn::Attribute to minimize breaking change impact

TedDriggs:
This change preemptively adds `syn::Attribute::meta_item(&self) -> Option<syn::MetaItem>` to 0.11.x. At the moment, it always returns `Some`, but when 0.12 is released it may begin returning `None` in cases that `MetaItem` parsing fails. This is intended to go out as 0.11.12.

### DIFF
--- a/src/attr.rs
+++ b/src/attr.rs
@@ -6,11 +6,18 @@ use std::iter;
 #[derive(Debug, Clone, Eq, PartialEq, Hash)]
 pub struct Attribute {
     pub style: AttrStyle,
+    #[deprecated(since = "0.11.12", note = "Not all attributes have to be well-formed meta items. \
+        Use `Attribute::meta_item` to attempt parsing instead")]
     pub value: MetaItem,
     pub is_sugared_doc: bool,
 }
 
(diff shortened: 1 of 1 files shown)


clarcharr:
I think that we should probably word this as `parse_meta_item`  to point out that it's not zero-cost.

TedDriggs:
Calling it `parse_meta_item` suggests to me that it would return some `Result` type, rather than an `Option`. I'm not attached to the name, but I'd like to write some code that won't break when 0.12 ships.

clarcharr:
I'd personally prefer a `Result` type to be honest.

dtolnay:
Thanks. This design is going to change as we revamp syn to operate on Macros 2.0-style tokens rather than strings, so the API currently in master probably won't end up in 0.12 after all.

TedDriggs:
@dtolnay is there anywhere I can plug in on that discussion?

dtolnay:
#142 and https://github.com/alexcrichton/syn and #rust-libs.

PR #135, opened 2017-05-20
Add more impls, methods to Ident.

clarcharr:
Adds:

* `Ident::as_str`
* `Ident::into_string`
* `Into<Cow<str>> for Ident`
* `Into<String> for Ident`
* `PartialOrd<T> for Ident where T: AsRef<str>`

### DIFF
--- a/src/ident.rs
+++ b/src/ident.rs
@@ -1,12 +1,21 @@
 use std::borrow::Cow;
+use std::cmp::Ordering;
 use std::fmt::{self, Display};
 
-#[derive(Debug, Clone, Eq, Hash, Ord, PartialOrd)]
+#[derive(Debug, Clone, Eq, Hash, Ord)]
 pub struct Ident(String);
 
 impl Ident {
     pub fn new<T: Into<Ident>>(t: T) -> Self {
         t.into()
+    }
+
+    pub fn as_str(&self) -> &str {
+        &self.0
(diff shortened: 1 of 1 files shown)


dtolnay:
Thanks!

1. Could you check on the build failure?
2. What is a use case for `PartialOrd<T> where T: AsRef<str>`? I have never seen an impl like that.
3. What is a use case for `Into<Cow<'a, str>>`?
4. The `Into` impls should be expressed as `From` impls if possible.

clarcharr:
1. Should be fixed!
2. I sort of consider it a completeness thing, although I was wrong in assuming that `String` implements `PartialOrd<str>`. I can remove it if it seems too out of place!
3. A lot of `String`-implemented traits have been implemented for `Cow` as well in the standard library, so I figured it'd make sense to add that here to allow bidirectional `Cow<str> <-> Ident` conversion.
4. I can't, unfortunately. Coherence rules would forbid implementing traits for libstd types.

dtolnay:
4. It should work. What was the error message?

dtolnay:
Ident is going to be replaced by [`proc_macro2::Symbol`](http://alexcrichton.com/proc-macro2/proc_macro2/struct.Symbol.html) in #142 which does not have these impls and methods, so I would prefer not to add them now.

PR #139, opened 2017-05-22
Add defaults to several types

TedDriggs:
The lack of `Default` impls on types such as `syn::Ident` prevents types with `Ident` fields from deriving `Default` themselves. This introduces considerable boilerplate for proc-macro crates such as my current project [darling](https://github.com/TedDriggs/darling).

This PR addresses that by providing defaults for a number of types within `syn`. Ideally, this would _not_ be blocked until 0.12 releases, and would go as a 0.11.x release.

### DIFF
--- a/src/data.rs
+++ b/src/data.rs
@@ -1,7 +1,7 @@
 use super::*;
 
 /// An enum variant.
-#[derive(Debug, Clone, Eq, PartialEq, Hash)]
+#[derive(Debug, Default, Clone, Eq, PartialEq, Hash)]
 pub struct Variant {
     /// Name of the variant.
     pub ident: Ident,
@@ -27,6 +27,12 @@ pub enum VariantData {
 
     /// Unit variant, e.g. `None`.
     Unit,
+}
+
+impl Default for VariantData {
+    fn default() -> Self {
+        VariantData::Unit
+    }
 }
 
 impl VariantData {
@@ -82,6 +88,12 @@ pub enum Visibility {
 
     /// Inherited, i.e. private.
     Inherited,
+}
+
+impl Default for Visibility {
+    fn default() -> Self {
+        Visibility::Inherited
+    }
 }
 
 #[cfg(feature = "parsing")]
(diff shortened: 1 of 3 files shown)


dtolnay:
But you would never want an Ident whose content is the empty string, right? Would Option\<Ident\> work?

TedDriggs:
@dtolnay `Option<Ident>` incorrectly gives the impression that it can be `None` when construction is complete. The issue I've been running into is this:

```rust
#[derive(FromDeriveInput)]
#[darling(attributes(from_variants))]
pub struct Container {
    pub into: bool,
    pub ident: syn::Ident,
    generics: syn::Generics,
    body: Body<Variant, Field>,
}
```

I'd like to allow the caller to specify `default` at the struct level and get serde-like fallbacks for values that aren't specified. However, `syn::Ident` doesn't implement `Default`, so `Container` has to implement it manually even though the value will be overridden moments after initialization. I've implemented `default` specifiers at the individual field level, but that adds a lot of verbosity and means there isn't one canonical default for the type, regardless of how it's made.

I've tried a couple other approaches:

## Separating AST and Attribute values
```rust
/// A general-purpose container for collecting data from a `syn::DeriveInput` instance.
#[derive(Debug, Clone, PartialEq, Eq)]
pub struct Container<A, B> {
    pub ident: syn::Ident,
    pub vis: syn::Visibility,
    pub generics: syn::Generics,
    pub attrs: A,
    pub body: B,
}
```

The problem with this approach is that the defined `Container` type is in my crate, not the caller's, so they can't easily add methods on without first defining a trait. The result doesn't feel very ergonomic.

## Support `from_ident` instead of `default`
This approach has the caller implement `From<syn::Ident>` for the type in question, and then invokes that instead of `Default::default()` to create the instance from which fallback values are taken. This approach _works_, but it's very verbose.

dtolnay:
If the default value is not meaningful and never meant to exist, it shouldn't be the default. Can you solve this in a different way? Maybe have FromDeriveInput generate the Default impl in a way that makes sense for you, or find a way to address this without an impl of Default.

PR #144, opened 2017-05-23
Rewrite the AST to be a bit more user-friendly

### LINKED ISSUE #136: User-friendly AST
Compare the AST from [strata-rust](https://github.com/shepmaster/strata-rust) and the one from syn:

```rust
pub enum Pattern {
    Character(PatternCharacter),
    Ident(PatternIdent),
    Number(PatternNumber),
    Range(PatternRange),
    Reference(PatternReference),
    String(PatternString),
    Struct(PatternStruct),
    Tuple(PatternTuple),
}
```

```rust
pub enum Pat {
    Wild,
    Ident(BindingMode, Ident, Option<Box<Pat>>),
    Struct(Path, Vec<FieldPat>, bool),
    TupleStruct(Path, Vec<Pat>, Option<usize>),
    Path(Option<QSelf>, Path),
    Tuple(Vec<Pat>, Option<usize>),
    Box(Box<Pat>),
    Ref(Box<Pat>, Mutability),
    Lit(Box<Expr>),
    Range(Box<Expr>, Box<Expr>),
    Slice(Vec<Pat>, Option<Box<Pat>>, Vec<Pat>),
    Mac(Mac),
}
```

One of these is more consistent and more predictable than the other.

alexcrichton:
This commit is a relatively large rewrite of the AST that `syn` exposes. The
main change is to expose enums-of-structs rather than
enums-with-huge-tuple-variants. The best example of this is `ItemKind::Fn` which
changed from:

    enum ItemKind {
        Fn(Box<FnDecl>, Unsafety, Constness, Option<Abi>, Generics, Box<Block>),
        ...
    }

to

    enum ItemKind {
        Fn(ItemFn),
        ...
    }

    struct ItemFn {
        decl: Box<FnDecl>,
        unsafety: Unsafety,
        constness: Constness,
        abi: Option<Abi>,
        generics: Generics,
        block: Box<Block>,
    }

This change serves a few purposes:

* It's now much easier to add fields to each variant of the ast, ast struct
  fields tend to be "by default ignored" in most contexts.
* It's much easier to document what each field is, as each field can have
  dedicated documentation.
* There's now canonicalized names for each field (the name of the field) which
  can help match `match` statements more consistent across a codebase.

A downside of this representation is that it can be a little more verbose to
work with in `match` statements and during constructions. Overall though I'd
feel at least that the readability improved significantly despite the extra
words required to do various operations.

Closes #136

### DIFF
--- a/src/aster/ty.rs
+++ b/src/aster/ty.rs
@@ -1,4 +1,5 @@
 use {Generics, Lifetime, MutTy, Mutability, Path, QSelf, Ty, TyParamBound};
+use {TyPath, TySlice, TyNever, TyInfer, TyTup, TyRptr, TyImplTrait};
 use aster::ident::ToIdent;
 use aster::invoke::{Invoke, Identity};
 use aster::lifetime::IntoLifetime;
@@ -36,11 +37,11 @@ impl<F> TyBuilder<F>
     }
 
     pub fn build_path(self, path: Path) -> F::Result {
-        self.build(Ty::Path(None, path))
+        self.build(Ty::Path(TyPath { qself: None, path: path }))
     }
 
     pub fn build_qpath(self, qself: QSelf, path: Path) -> F::Result {
-        self.build(Ty::Path(Some(qself), path))
+        self.build(Ty::Path(TyPath { qself: Some(qself), path: path }))
     }
 
     pub fn path(self) -> PathBuilder<TyPathBuilder<F>> {
@@ -115,7 +116,7 @@ impl<F> TyBuilder<F>
     }
 
     pub fn build_slice(self, ty: Ty) -> F::Result {
-        self.build(Ty::Slice(Box::new(ty)))
+        self.build(Ty::Slice(TySlice { ty: Box::new(ty) }))
     }
 
     pub fn slice(self) -> TyBuilder<TySliceBuilder<F>> {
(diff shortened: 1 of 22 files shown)


alexcrichton:
Note that tons of the names here are bikesheddable, I wouldn't mind tweaking anything!

dtolnay:
Amazing!

How would you feel about putting these tuple variants in modules? For example `expr::IfLet` instead of `ExprIfLet`. That would help avoid the even-worse wall of types when you open the documentation. I'll merge this quickly and we can follow up separately if you agree.

alexcrichton:
Sounds great to me!

alexcrichton:
Ah so I've discovered that if we do `IfLet` as a struct name then you can't import both `ExprKind::*` and `expr::*` unfortunately, leaving us with a situation that chooses between a number of the following in a match statement:

```rust
// currently proposed
IfLet(ExprIfLet { ... }) => 

// Drop "Expr" prefix
ExprKind::IfLet(IfLet { ... }) => 
// or
IfLet(expr::IfLet { ... }) => 

// another random prefix?
IfLet(EIfLet { ... }) => 
```

dtolnay:
Let's keep the current approach for enums.

The build failure looks like TyPtr not printing "const".

alexcrichton:
Alright I think that'll do it!

alexcrichton:
Ah looks like I missed a few things when testing in 1.12.1, but now it's getting into the realm of "may not be super easy to fix". Looks like this'd bump to 1.15.0+

dtolnay:
That's fine. I'll remove those Travis builds.

dtolnay:
Up to Rust 1.15.1 in d9e2dc905ceedcd80ef5acffc990dab7b1583f7b.

alexcrichton:
👍

PR #145, opened 2017-05-23
Move all #[derive] impls behind Cargo feature gates

### LINKED ISSUE #143: Put #[derive] traits behind cfg
Right now with no-default-features it takes 3 seconds to compile and generates 7677 lines of code for all the junk that we derive on the AST. Most of that is never used in your typical proc macro library and should be off by default.

alexcrichton:
This commit moves all #[derive] annotations behind Cargo feature gates to add
the ability to strip them all out entirely. The `Clone` and `Copy` impls
continue to be enabled by default as they tend to be mega useful but other
equality/hash/debug impls are all default behind the `extra-impls` gate.

This commit, on my computer, has the following timings:

| features                      | before  | after
|-------------------------------|---------|------
| default                       | 3.67    | 2.96
| *none*                        | 1.78    | 0.49
| {printing, parsing}           | 3.71    | 2.57
| default + {full}              | 8.50    | 6.31
| {full}                        | 3.53    | 0.70
| {full, printing, parsing}     | 8.10    | 5.29

Closes #143

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,11 +38,14 @@ script:
           cargo build --features full &&
           cargo build --features 'visit fold' &&
           cargo build --features 'full visit fold' &&
+          cargo build --no-default-features --features 'full parsing printing' &&
+          cargo build --no-default-features --features 'visit fold parsing printing' &&
+          cargo build --no-default-features --features 'full visit fold parsing printing' &&
           (cd synom && cargo test)
           ;;
         test)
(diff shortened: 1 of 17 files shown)


dtolnay:
I am looking into what serde_derive needs from these impls. It would be great to have serde_derive not need the full set.

dtolnay:
Yes, we can move to only Clone with some easy changes.

Issue #142, opened 2017-05-22
Rewrite to operate on tokens


dtolnay:
Using https://github.com/rust-lang/rust/pull/40939 and [`proc-macro2`](https://github.com/alexcrichton/proc-macro2). **@​alexcrichton** is working on this.

mystor:
I'm helping @alexcrichton with modifying the parser to run on TokenStream objects. My current unrebased branch is https://github.com/mystor/syn/tree/proc-macro2.

dtolnay:
I love both of you.

@mystor question about `flatten_stream` - it looks like this would be lossy right? If you have a TokenStream containing `$a * 3` where the value of $a is `1 + 2`, this would turn it into something that is indistinguishable from `1 + 2 * 3`. (Not that we handle operator precedence anyway currently, but we will want to.) Any sense of how much harder it would be to support that correctly?

mystor:
@dtolnay Yes, `flatten_stream` is currently lossy, which is quite unfortunate. I don't know off of the top of my head what the best way to handle the `None` delimiter sequences.

As you can probably tell from my current WIP branch, I was trying to do the minimal number of changes to the rest of `syn` to get the parser to work built on top of `proc-macro2`. I imagine that any attempt to get `None`-delimited sequences working with precedence will require a reworking of that entire part of the parser, as well as a change to the `synom` input format, which I figured we would probably have to do separately (although potentially before `0.12` is released, depending on how long it takes to get the real `proc-macro` into nightly).

PR #146, opened 2017-05-26
Rewrite tokenization with `proc-macro2` tokens

alexcrichton:
This ended up being a bit larger of a commit than I intended! I imagine that
this'll be one of the larger of the commits working towards #142. The purpose of
this commit is to use an updated version of the `quote` crate which doesn't work
with strings but rather works with tokens form the `proc-macro2` crate. The
`proc-macro2` crate itself is based on the proposed API for `proc_macro` itself,
and will continue to mirror it. The hope is that we'll flip an easy switch
eventually to use compiler tokens, whereas for now we'll stick to string parsing
at the lowest layer.

The largest change here is the addition of span information to the AST. Building
on the previous PRs to refactor the AST this makes it relatively easy from a
user perspective to digest and use the AST still, it's just a few extra fields
on the side. The fallout from this was then quite large throughout the
`printing` feature of the crate. The `parsing`, `fold`, and `visit` features
then followed suit to get updated as well.

This commit also changes the the semantics of the AST somewhat as well.
Previously it was inferred what tokens should be printed, for example if you
have a closure argument `syn` would automatically not print the colon in `a: b`
if the type listed was "infer this type". Now the colon is a separate field and
must be in sync with the type listed as the colon/type will be printed
unconditionally (emitting no output if both are `None`).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,17 +13,22 @@ include = ["Cargo.toml", "src/**/*.rs", "README.md", "LICENSE-APACHE", "LICENSE-
 default = ["parsing", "printing", "clone-impls"]
 aster = []
 full = []
-parsing = ["unicode-xid", "synom"]
-printing = ["quote"]
+parsing = ["unicode-xid", "synom/parsing"]
+printing = ["quote", "synom/printing"]
 visit = []
 fold = []
 clone-impls = []
 extra-traits = []
 
 [dependencies]
-quote = { version = "0.3.7", optional = true }
+proc-macro2 = { git = 'https://github.com/alexcrichton/proc-macro2' }
 unicode-xid = { version = "0.0.4", optional = true }
-synom = { version = "0.11", path = "synom", optional = true }
+synom = { version = "0.11", path = "synom", default-features = true }
+
+[dependencies.quote]
+git = 'https://github.com/alexcrichton/quote'
+branch = 'new-tokens'
+optional = true
 
 [dev-dependencies]
 syntex_pos = "0.58"
--- a/src/aster/generics.rs
+++ b/src/aster/generics.rs
@@ -5,6 +5,8 @@ use aster::path::IntoPath;
 use aster::ty_param::TyParamBuilder;
 use aster::where_predicate::WherePredicateBuilder;
 
+use delimited::Delimited;
+
 pub struct GenericsBuilder<F = Identity> {
     callback: F,
     lifetimes: Vec<LifetimeDef>,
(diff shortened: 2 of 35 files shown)


alexcrichton:
Alright so this is a bit of a monster PR! I'm hoping this is the final version of the AST before we close out #142. Next steps I can see are:

* [ ] Review the [`proc-macro2` crate](https://github.com/alexcrichton/proc-macro2)
* [ ] Publish the `proc-macro2` crate
* [ ] Review the [changes to the `quote` crate](https://github.com/dtolnay/quote/compare/master...alexcrichton:new-tokens)
* [ ] Publish the `quote` crate
* [ ] Review this PR
* [ ] Update serde_derive to work with this PR before merging perhaps?
* [ ] Merge

So far I'm pretty happy with how this turned out, it ended up being pretty easy to implement parsing, `ToTokens`, visiting, and folding after this change. I don't *think* there's redundant information (only what's needed), but an extra pair of eyes is always helpful!

alexcrichton:
I'll also note that I ended up diverging somewhat from the proposed `proc_macro` API. I tried to stick as closely as possible to the current revision in https://github.com/rust-lang/rust/pull/40939 as possible, but to get all the tests passing in this repo (especially the round-trip ones) I had to significantly [bolster the `Literal` API](https://github.com/alexcrichton/proc-macro2/blob/2d0cf0bcc36d578f55e3e2a8ea71fb147921ef85/src/lib.rs#L163-L191) to meet the various needs.

I imagine we'll solve all these issues before stabilizing `proc_macro`, however.

alexcrichton:
Also note that this does not parse from tokens yet, this only updates tokenization. All parser code is "untouched" minus the updates necessary to get this new AST being generated.

dtolnay:
This is pretty clearly the future so I am comfortable merging this and following up with any necessary tweaks separately, rather than futzing with dependent PRs.

My only concern is I would like to have the real proc-macro API from https://github.com/rust-lang/rust/pull/40939 in better shape before publishing either of proc-macro2 or syn. Experimenting with a bolstered API is fine but when publishing a syn release I would like to make sure that proc-macro2 has "stable" and "unstable" modules that are perfectly identical.

alexcrichton:
Yeah sounds good to me. If it's ok I think we should hold off on publishing proc-macro2 and the next version of this crate until the change to rustc happens. I think that the current API is *sufficient* albeit very difficult to work around, but we could make do if need be. 

My current thinking is that when the `proc_macro` API lands on nightly we'll freeze the `stable` module and the interface of the `proc_macro2` crate to look exactly like that. As the `proc_macro` API changes (if it does) we'll update the `unstable` module to do a translation between the old and the new, maintaining API compatibility of the `proc_macro2` crate itself. Then finally once we get closer to stabilization I figure we can release a new breaking change.

PR #149, opened 2017-05-31
Use nom macro to declare parse method inside Synom trait

dtolnay:
In the `do_parse!(input, ...)` way of invoking nom-style macros, the "input" argument seems like a nom implementation detail that users aren't supposed to need to write.


dtolnay:
I filed #150 to follow up in other places.

alexcrichton:
*mind blown*

Issue #151, opened 2017-06-01
Consider switching `synom` from `IResult<I, O>` to `Result<(I, O), ...>`


mystor:
Right now we use a custom result type for `synom`. In `nom` this was done to support the non-error `Incomplete` variant, which we don't use in `syn`. We don't have that need in synom, as we only have the successful and unsuccessful parse conditions. We should consider replacing our `enum IResult` with:

```rust
type Result<'a, O> = Result<(&'a [TokenStream], O), ParseError>
```

We could also make a new type instead of using the tuple on the Ok() side of the result. If we do that I would probably want to add `fn parse_ok<O>(rest: &[TokenStream], val: O) -> Result<O>` and `fn parse_err<O>() -> Result<O>` helper methods for ergonomic reasons.

For #148 I will probably change the input type from `&[TokenStream]` to a `Cursor` type.

dtolnay:
I agree. This is long overdue.

mystor:
I'll do this before #148, as it'll make that patch easier.

dtolnay:
You are only working on this for syn/synom right? Proc-macro2 might benefit from this too https://github.com/alexcrichton/proc-macro2/issues/12.

mystor:
Yes, I just started writing the patch for `syn/synom`. I wasn't intending to make the change to `proc-macro2`, but we should do it there as well.

dtolnay:
Fixed in #152.

Issue #148, opened 2017-05-31
Handle precidence and correctly parsing `None`-delimited spans


mystor:
We currently don't handle precedence when parsing expressions on `master`. We should make sure that we get expression precedence correct and correctly handle spans generated by macros.

I'm working on this right now. Filing this bug to track the work.

PR #152, opened 2017-06-01
Switch from `IResult<I, O>` to `Result<(I, O), E>`

### LINKED ISSUE #150: Use nom macro to declare parse method inside Synom trait
As in #149, but for all the other impls.

In the `do_parse!(input, ...)` way of invoking nom-style macros, the "input" argument seems like a nom implementation detail that users aren't supposed to need to write.

mystor:
Fixes #150, #151 

The bulk of this patch is actually #150, as I needed to rewrite every `parse` signature to use the new types, and it was easy enough to just convert them into invocations of `named!` every time.

I also introduced a `Cursor` type alias in `synom` which I use instead of `&[TokenTree]` as the parser input type. I intend in #148 to change the `Cursor` type to be something other than `&[TokenTree]` so this will make my life easier in the future.

I use the type alias of `PResult<'a, O>` to refer to parser results (of the form written above).

### DIFF
--- a/src/attr.rs
+++ b/src/attr.rs
@@ -292,7 +292,7 @@ impl<'a, T> FilterAttrs<'a> for T
 #[cfg(feature = "parsing")]
 pub mod parsing {
     use super::*;
-    use synom::IResult;
+    use synom::{PResult, Cursor, parse_error};
     use synom::tokens::*;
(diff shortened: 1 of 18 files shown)


mystor:
@alexcrichton I fixed the test failure - should pass now :)

PR #155, opened 2017-06-03
Switch away from using `&[TokenStream]` as the input type

### LINKED ISSUE #148: Handle precidence and correctly parsing `None`-delimited spans

mystor:
The new input type for parsers in `syn` with this patch is the type `synom::Cursor`. This is an unsafely implemented type, which holds raw pointers into a formatted `SynomBuffer` object. This change allows for a few things:

1. We no longer need to allocate a `Vec` every time we want to enter a delimited section. The cursor just moves into the correct part of the backing buffer, and
2. This sets us up to fix #148 as we no longer erase `None`-delimited spans from our input, and can both ignore them when parsing most inputs, while still being able to use them for expression parsing.

The new `synom::cursor` module in the first patch adds some unsafe code in order to implement the more efficient cursor. It would probably be a good idea to double-check that I'm not violating any of my constraints in that module, and that the API I expose is actually safe.

### DIFF
--- a/synom/src/cursor.rs
+++ b/synom/src/cursor.rs
@@ -1,0 +1,358 @@
+//! This module defines a cheaply-copyable cursor into a TokenStream's data.
+//!
+//! It does this by copying the data into a stably-addressed structured buffer,
+//! and holding raw pointers into that buffer to allow walking through delimited
+//! sequences cheaply.
+//!
+//! This module is heavily commented as it contains the only unsafe code in
+//! `syn`, and caution should be made when editing it. It provides a safe
+//! interface, but is fragile internally.
+
+use proc_macro2::*;
+
+use std::ptr;
+use std::fmt;
+use std::marker::PhantomData;
+
+/// Internal type which is used instead of `TokenTree` to represent a single
+/// `TokenTree` within a `SynomBuffer`.
+#[derive(Debug)]
+enum Entry {
+    /// Mimicing types from proc-macro.
+    Sequence(Span, Delimiter, SynomBuffer),
+    Word(Span, Symbol),
+    Op(Span, char, OpKind),
+    Literal(Span, Literal),
+    /// End entries contain a raw pointer to the entry from the containing
+    /// TokenTree.
+    End(*const Entry),
+}
+
+// A `Sync` impl is necessary here because we need a global `Entry` object for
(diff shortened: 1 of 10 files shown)


alexcrichton:
👍

synom/src/cursor.rs
```diff
@@ -0,0 +1,357 @@
...
+            _ => None
+        }
+    }
+
+    /// If the cursor is pointing at a Word, return it and a cursor pointing at
+    /// the next `TokenTree`.
+    pub fn word(mut self) -> Option<(Cursor<'a>, Span, Symbol)> {
+        self.ignore_none();
+        match *self.entry() {
+            Entry::Word(span, ref sym) => {
```
dtolnay: Symbol implements Copy so we shouldn't need ref here.

synom/src/cursor.rs
```diff
@@ -0,0 +1,357 @@
...
+    Op(Span, char, OpKind),
+    Literal(Span, Literal),
+    /// End entries contain a raw pointer to the entry from the containing TokenTree
+    End(*const Entry),
+}
+
+// A sync impl is necessary here because we need a global Entry object for the
+// empty sequence case. This impl is requied as `*const Entry` is neither `Send`
+// nor `Sync` by default.
+unsafe impl Sync for Entry {}
```
dtolnay: Should be fine to Send as well, right? Everything in here is immutable.

Let's also add Send and Sync for Cursor which is basically &Entry.

dtolnay:
With 100% seriousness, cursor.rs is my favorite Rust code I have ever read. Merge when ready.

synom/src/cursor.rs
dtolnay: This comment was wrong - https://github.com/alexcrichton/proc-macro2/issues/24. We will need to keep Sync for Entry because of the static (which we know does not contain a Symbol) but that is fine as long as nothing public is Sync.

mystor: I'm fixing this in #157. We can't make `Entry` Sync because that would make `SynomBuffer` sync implicitly, which we definitely don't want. I added a custom wrapper type instead.

PR #165, opened 2017-06-04
Pare down the Synom trait

### LINKED ISSUE #164: Rename syn::Crate
The name comes from libsyntax's ast::Crate but it is misleading for us because I think libsyntax parses and inlines all the modules that are referenced by the code, while we parse a single file only.

dtolnay:
I would like to make it clearer that parsing a string is second-class functionality compared to parsing tokens.

Fixes #164.


dtolnay:
This is somewhat confusing in that this public API uses `proc_macro::TokenStream` while synom's public API uses `proc_macro2::TokenStream`.

src/lib.rs
```diff
@@ -105,45 +106,168 @@ pub mod visit;
...
+/// }
+/// #
+/// # quick_main!(run);
+/// ```
+#[cfg(feature = "parsing")]
+pub fn parse_str<T: Synom>(s: &str) -> Result<T, ParseError> {
+    parse(s.parse()?)
+}
+
+// FIXME the name parse_file makes it sound like you might pass in a path to a
```
mystor: If we decide to add automatic module discovery at some point we would want to have a `parse_file` where you do pass in a path. I'd consider calling it `parse_mod`?

src/lib.rs
```diff
@@ -105,45 +106,168 @@ pub mod visit;
...
+///     let expanded = quote! {
+///         /* ... */
+///     };
+///
+///     // Parse back to a token stream and return it
+///     expanded.parse().unwrap()
+/// }
+/// ```
+#[cfg(feature = "parsing")]
+pub fn parse<T: Synom>(tokens: proc_macro::TokenStream) -> Result<T, ParseError> {
```
mystor: Can we make this take a trait object which is implemented both on `proc_macro::TokenStream` and `proc_macro2::TokenStream` so that it acts less confusingly? it would have a method to turn it into a `proc_macro2::TokenStream`. Perhaps `Into<proc_macro2::TokenStream>`?

That way we just DTRT, which could be really nice. This is especially the case when dealing with parsing code from outside of a procedural macro, where `parse<T>` wouldn't be usable, as you cannot create a `proc_macro::TokenStream` from outside of a procedural macro right now without triggering a panic.

tests/test_round_trip.rs
```diff
@@ -111,7 +111,7 @@ fn test_round_trip() {
         file.read_to_string(&mut content).unwrap();
 
         let start = Instant::now();
-        let (krate, elapsed) = match content.parse::<syn::Crate>() {
+        let (krate, elapsed) = match syn::parse_str::<syn::File>(&content) {
```
mystor: Does this actually work? Shouldn't you use `parse_file` here to handle BOMs and shebangs?

src/lib.rs
dtolnay: I don't love parse_mod because modules are not always files.

```rust
mod mystor {
    /* ... */
}
```

Really I would prefer to get rid of this function and move this logic into the test_round_trip code. We only have this because some of the Rust test suite exercises these obscure features. Nobody needs us to have this in the public API. What do you think?

src/lib.rs
dtolnay: When would you ever need to parse something outside of a procedural macro where `syn::parse_str` would not be sufficient? That sounds like an advanced enough use case that the user should be comfortable using the Synom trait directly.

I'm not against using an input trait like you said but I'd like to make sure it is solving a real problem - functions where both the input and output are type parameters are substantially more complicated to grok compared to just one type parameter.

src/lib.rs
mystor: I do have a consumer of `syn` which directly parses raw rust files, and I know of a few others, and I would generally want to be able to handle munching off the BOM and shebang if present from these files before parsing them, but I don't think it's unreasonable to pass off that responsibility to the consumer crate (or potentially a helper crate) rather than doing it in syn. It seems like a tangential responsibility to me.

I would just remove this and only expose some form of `parse_mod` which takes in a `TokenStream`, and then in the `test_round_trip` we could just throw away the BOM and shebang if present.

src/lib.rs
mystor: Fair - I would probably be the `Synom` trait directly in that case anyway. Might as well leave this at directly taking `proc_macro::TokenStream`.

dtolnay:
@alexcrichton do you think anything here is worth picking up for 0.12 or do we follow up later?

alexcrichton:
I think this is worth doing, I'll look to rebasing/fixing.

alexcrichton:
Ok so to make this testable I had to change this to using `Into<proc_macro2::TokenStream>` instead of `proc_macro::TokenStream`. The latter requires the context of the compiler to actually be created, effectively making `proc_macro` impossible to use in unit tests. (at least as constructed today)

src/fold.rs
```diff
@@ -564,13 +564,13 @@ pub fn noop_fold_mac<F: ?Sized + Folder>(folder: &mut F, mac: Mac) -> Mac {
...
 #[cfg(feature = "full")]
-pub fn noop_fold_crate<F: ?Sized + Folder>(folder: &mut F,
-                                           krate: Crate)
-                                           -> Crate {
-    Crate {
-        attrs: krate.attrs.lift(|a| folder.fold_attribute(a)),
-        items: krate.items.lift(|i| folder.fold_item(i)),
-        ..krate
+pub fn noop_fold_file<F: ?Sized + Folder>(folder: &mut F,
+                                           file: File)
```
mystor: nit: Indentation

synom/src/lib.rs
```diff
@@ -63,38 +61,6 @@ pub trait Synom: Sized {
...
-            Err(ref err) => format!("{} while parsing {}", err.description(), descr),
-        };
-        Err(ParseError(Some(err)))
-    }
-
-    fn parse_str_all(input: &str) -> Result<Self, ParseError> {
-        Self::parse_all(input.parse()?)
-    }
-
-    fn parse_all_unwrap(input: TokenStream) -> Self {
```
mystor: I suppose we currently aren't providing an API like this, and instead encouraging people to use `syn::parse::<T>().unwrap()`?

tests/test_precedence.rs
```diff
@@ -95,9 +95,9 @@ fn test_rustc_precedence() {
         let mut content = String::new();
         file.read_to_string(&mut content).unwrap();
 
-        let (l_passed, l_failed) = match content.parse::<syn::Crate>() {
-            Ok(krate) => {
-                let exprs = collect_exprs(krate);
+        let (l_passed, l_failed) = match syn::parse_str::<syn::File>(&content) {
```
mystor: You'll need to update this to use the BOM and shebang-handling `syn::parse_file` function instead. Some of these input strings contain those characters.

tests/test_round_trip.rs
mystor: I already mentioned this, but this method probably needs to use `parse_file` as well.

src/lib.rs
```diff
@@ -100,45 +100,171 @@ pub mod visit;
...
+///     let expanded = quote! {
+///         /* ... */
+///     };
+///
+///     // Parse back to a token stream and return it
+///     expanded.parse().unwrap()
+/// }
+/// ```
+#[cfg(feature = "parsing")]
+pub fn parse<T, U>(tokens: U) -> Result<T, ParseError>
```
mystor: It's a tad unfortunate that we can't write `syn::parse::<Expr>` or something similar anymore, as we have a second type parameter, and instead need to write `syn::parse::<Expr, _>`.  I can't think of a better option though, so :-/.

dtolnay: > Ok so to make this testable I had to change this to using `Into<proc_macro2::TokenStream>` instead of `proc_macro::TokenStream`. The latter requires the context of the compiler to actually be created, effectively making `proc_macro` impossible to use in unit tests. (at least as constructed today)

If this is really only motivated by unit tests as Alex wrote, I would prefer to have a hidden function for the tests to use and stick with `proc_macro::TokenStream` for the public one. `syn::parse::<Expr, _>` is not nice.

synom/src/lib.rs
alexcrichton: Oh I was just following the previous changes, I don't particularly mind either way. Would be backwards compatible to add later though!

src/lib.rs
alexcrichton: Good points! I've reverted to take `proc_macro::TokenStream` here and got tests working otherwise.

mystor: In that case, I agree that just having it accept a proc_macro::TokenStream is nicer. 

Perhaps we could have an (optionally `#[doc(hidden)]`) `parse2` method which takes a `proc_macro2::TokenStream` as an argument?

Making it such that we can parse a `proc_macro::TokenStream` when not in the compiler does seem pretty important for the ergonomics of this API though. Perhaps we want to make `parse2` public until we can get this feature, and then deprecate and remove.

alexcrichton: Oh for now I opted to avoid a public API here, but if y'all think we should have one I can add that in too.

synom/src/lib.rs
dtolnay: I would hold off until it can actually behave better than `syn::parse::<T>().unwrap()`. We don't have `FromStr::from_str_unwrap` for example.

src/lib.rs
dtolnay: Let's wait until someone needs `parse2` and evaluate whether it is the best solution for what they are doing. As I wrote above:

> When would you ever need to parse something outside of a procedural macro where `syn::parse_str` would not be sufficient? That sounds like an advanced enough use case that the user should be comfortable using the Synom trait directly.

alexcrichton:
Ok looks like we're all good! I'm gonna merge, but we've still got time to tweak before serde_derive is updated

PR #168, opened 2017-06-05
Handle precedence when parsing expressions

### LINKED ISSUE #163: Parse Respecting Operator Precedence
#148 was closed when #155 was merged, despite precedence not being implemented yet (that patch only implements the new cursor type). This bug is for the second half of that work.

This is important for the correctness of round-tripping `None`-delimited spans, as they effect how expressions should be parsed, and messing them up could be problematic for the correctness of procedural macros built on `syn`.

### LINKED ISSUE #166: Fix `fold` and `visit` features
`master` doesn't currently build when the `fold` or `visit` features are enabled. We should fix this.

mystor:
This should fix #163.

This changes a bunch of stuff, adding a few new debugging tools which I found useful for fixing the test failures while writing these patches, some changes to `Ty`'s parsing, a ton of `Expr` parsing changes etc.

I have not written good tests for the precedence parsing yet. I have an idea of how to do it well, comparing us against `syntex` on rustc's source, like `test_round_trip` does, but I'm going to use `fold` to implement it, so I'll have to fix #166 before I can do that. This doesn't regress any tests though (it actually fixes a few of them!), and we aren't shipping this version of `syn` yet, so I would like to land this patch, and add the tests in a follow-up :).

### DIFF
--- a/src/expr.rs
+++ b/src/expr.rs
@@ -29,7 +29,7 @@ ast_enum_of_structs! {
             pub box_token: tokens::Box_,
         }),
 
-        /// E.g. 'place <- val'.
+        /// E.g. 'place <- val' or `in place { val }`.
         pub InPlace(ExprInPlace {
             pub place: Box<Expr>,
             pub value: Box<Expr>,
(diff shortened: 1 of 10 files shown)


dtolnay:
I don't have time to look at this tonight and I am starting a new job tomorrow so I may not have time all week. You and @alexcrichton are on your own! Also it would be great if one of you could take over #165.

alexcrichton:
Looks great to me! Follow-up tests sound good too

PR #169, opened 2017-06-05
Fix Visit and Fold to work on master

### LINKED ISSUE #166: Fix `fold` and `visit` features

mystor:
I did the minimum amount of work to get them working again. It would probably be worth going over them again to make sure we aren't missing important fields.

Fixes #166.

NOTE: This is built on top of #168, so there are a ton of patches which aren't relevant here until that is merged.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,16 +36,20 @@ script:
           cargo build --no-default-features &&
           cargo build &&
           cargo build --features full &&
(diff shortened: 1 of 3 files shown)


alexcrichton:
@mystor want to merge after a rebase?

Issue #170, opened 2017-06-06
Treat toplevel block-suffixed expressions as statements


mystor:
Currently we parse the expression:

```rust
for i in x {
    ()
}
*ptr = 10;
```
as
```rust
(for (i) in (x) {
   ()
} * ptr) = 10;
```
instead of the correct:
```rust
(for (i) in (x) {
    ()
});
((*ptr) = (10));
```

When appearing at the statement level, block-suffixed expressions shouldn't require a semicolon to be terminated, and should in fact implicitly terminate.

mystor:
Fixed by #172

PR #172, opened 2017-06-07
Test our expression parser correctly parses precedence

mystor:
This adds a test which, like test_round_trip, tests that our parser correctly parses the precedence of each expression.

It has a few caveats, which are documented in the respective parsers, which mostly exist because trying to compare ASTs parsed by two completely different parsers is a little tricky sometimes.

I also factored out some of the code from test_round_trip into a separate shared module, so that both test_round_trip and test_precedence can use it.

Finally I found some bugs in syn while writing this test (unsurprisingly), so I fixed those as well.

### DIFF
--- a/src/expr.rs
+++ b/src/expr.rs
@@ -1098,6 +1098,26 @@ pub mod parsing {
         syn!(ExprRepeat) => { ExprKind::Repeat }
     ));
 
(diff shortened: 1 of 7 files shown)


mystor:
This fixes #170

PR #173, opened 2017-06-07
Remove the MacroInput type alias

### LINKED ISSUE #167: Remove `MacroInput`
We're doing major breaking changes to the API with the `proc_macro2` changes. We should be able to remove the old type alias `MacroInput`, in favor of requiring consumers to change to `DeriveInput`. This will be a small change compared to the rest of the changes they'll have to make.

mystor:
Fixes #167.

@dtolnay How do you feel about removing this type alias?


dtolnay:
Please also make a PR to update https://doc.rust-lang.org/book/procedural-macros.html.

PR #175, opened 2017-06-08
Remove `ConstExpr` in favor of a limited `Expr`

mystor:
This regresses default-features all-in build times by about 0.1-0.3s on my laptop from ~7.4-7.6s (current master) to ~7.76s (mystor/constexpr - the new times seem more consistent?). In the past when I measured the regression it seemed more significant. I think this added overhead is mostly due to our now-correct precedence handling when parsing binary constant expressions. I figure this isn't too big of a build time regression but would like @dtolnay's opinion.

The advantage of this patch set is that it simplifies the interface to consistently use the same expression everywhere, and enables code re-use for the expression parser, by eliminating the chance that the `Expr` and `ConstExpr` parsers get out of sync.


dtolnay:
That seems fine. The Expr / ConstExpr split is not worth it to save 0.1-0.3 seconds.

PR #183, opened 2017-07-08
Remove `rust` as a submodule

alexcrichton:
This removes the `tests/rust` submodule and instead moves to lazily cloning it
during tests. This helps work around the fact that git dependencies on `syn`
will pull in all submodules, and rust takes forever to clone. In short, the
intention here is to make git dependencies on `syn` bearable to avoid Cargo
auto-cloning submodules.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 target
 Cargo.lock
+tests/rust
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,0 @@
-[submodule "tests/rust"]
-	path = tests/rust
-	url = https://github.com/rust-lang/rust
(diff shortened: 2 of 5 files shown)


mystor:
This is a very good idea :). 

I was going to suggest only shallow-cloning the repository so it clones faster, but it seems like that's a tad tricky to do as you can't shallow-clone based on a commit.

Perhaps we could skip git cloning at all, and just download a zip with the repo at that commit from github (like https://github.com/rust-lang/rust/archive/ddc5d7bd4b9ea3e8a8ccf82cb443e950be311d61.zip), unzip that, and use it?

alexcrichton:
This is typically only run on builders so I figured it wasn't worth the 30 or so seconds saved for going out of our way to get shallow clones working.

PR #184, opened 2017-07-08
Make `ToTokens` more lenient when generating tokens for invalid ASTs

### LINKED ISSUE #177: Somehow prevent the creation of invalid ASTs
This was mentioned as a pitfall in https://github.com/serde-rs/serde/pull/980#issuecomment-313534403.

Alternatively, we could somehow provide an API for correcting an AST into a valid AST which we could potentially run before tokenifying each AST node.

Some examples of things which would be nice to prevent include:
1. Missing tokens in structs which are necessary for the output
2. Missing parentheses or none-delimited spans which make the generated tokens parse incorrectly.
3. Maybe more things?

mystor:
This is my first pass at making our `ToTokens` implementations more lenient when it comes to generating tokens for invalid ASTs. I have some comments in the patch in the places which I both a) noticed as I read through all of our `ToTokens` implementations, and b) didn't fix.

This patch contains a few large if conditionals, and I'm really not sure how to nicely format those in the style used by syn, so I just... gave up? I tried running rustfmt for inspiration, but the style `syn` uses is so far from rustfmt's suggestions right now that it felt out of place. I'd love suggestions as to how to fix them to not be so ugly.

Fixes #177

### DIFF
--- a/src/ty.rs
+++ b/src/ty.rs
@@ -840,7 +840,7 @@ mod printing {
             match self.ty.mutability {
                 Mutability::Mutable(ref tok) => tok.to_tokens(tokens),
                 Mutability::Immutable => {
-                    self.const_token.unwrap_or_default().to_tokens(tokens);
+                    TokensOrDefault(&self.const_token).to_tokens(tokens);
                 }
             }
             self.ty.ty.to_tokens(tokens);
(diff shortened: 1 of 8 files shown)


src/ty.rs
```diff
@@ -984,7 +998,21 @@ mod printing {
             self.turbofish.to_tokens(tokens);
             self.lt_token.to_tokens(tokens);
             self.lifetimes.to_tokens(tokens);
+            if !self.lifetimes.empty_or_trailing() && !self.types.is_empty() {
+                tokens::Comma::default().to_tokens(tokens);
+            }
             self.types.to_tokens(tokens);
+            if (
```
mystor: This conditional is the one which I gave up the most on :-/. Not sure how to make it not-ugly.

mystor:
Thanks for merging this @alexcrichton. I didn't have enough time to fix it up :).

PR #186, opened 2017-07-10
Support for `yield` expressions

alexcrichton:
This adds support for `yield` expressions as added in https://github.com/rust-lang/rust/pull/43076

### DIFF
--- a/synom/src/tokens.rs
+++ b/synom/src/tokens.rs
@@ -190,6 +190,7 @@ tokens! {
         (pub struct Use                     => "use"),
         (pub struct Where                   => "where"),
         (pub struct While                   => "while"),
+        (pub struct Yield                   => "yield"),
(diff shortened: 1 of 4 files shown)


synom/src/tokens.rs
```diff
@@ -190,6 +190,7 @@ tokens! {
         (pub struct Use                     => "use"),
         (pub struct Where                   => "where"),
         (pub struct While                   => "while"),
+        (pub struct Yield                   => "yield"),
```
mystor: Should we consider adding more tokens here for other keywords reserved for future use? It seems weird that we had to add this to our tokens list when it was already a reserved keyword in rust.

alexcrichton: Seems fine to me yeah!

PR #193, opened 2017-08-28
Update the rust submodule we're testing

alexcrichton:
Was playing around with async/await and ran into syntax that syn can't parse. Took at stab in the dark an prayed that updating the `rust` submodule and fixing tests would fix my original test. It didn't. Figured this'd be useful regardless!


tests/test_precedence.rs
```diff
@@ -91,6 +89,9 @@ fn test_rustc_precedence() {
         if path.is_dir() {
             continue;
         }
+        if path.to_str().unwrap().ends_with("optional_comma_in_match_arm.rs") {
```
mystor: nit: Can you add a comment here explaining why this test fails?

src/main.rs
```diff
@@ -0,0 +1,11 @@
+extern crate syn;
+
+use std::io::{self, Read};
+
+fn main() {
+    // let mut input = String::new();
+    // io::stdin().read_to_string(&mut input).unwrap();
+    //
+    // syn::parse_str::<syn::File>(&input).unwrap();
```
mystor: Please remove this file, I don't think it's necessary.

src/expr.rs
```diff
@@ -624,11 +624,20 @@ ast_enum! {
 
 #[cfg(any(feature = "parsing", feature = "printing"))]
 #[cfg(feature = "full")]
-fn arm_requires_comma(arm: &Arm) -> bool {
-    if let ExprKind::Block(ExprBlock { unsafety: Unsafety::Normal, .. }) = arm.body.node {
-        false
-    } else {
-        true
+fn arm_expr_requires_comma(expr: &Expr) -> bool {
```
dtolnay: I don't think this is correct -- the rule for expressions that need a semicolon in order to be a statement is different from the rule for expressions that need a comma in order to be a match arm. The link is for the former. I haven't tried it but does this change make syn erroneously parse [this code](https://play.rust-lang.org/?gist=5cf023825e7eeffa1ff5ac42f293c60a)?

```rust
fn main() {
    match 0 {
        1 => loop {}
        _ => {}
    }
}
```

dtolnay: Oh this changed in nightly! Never mind.

src/main.rs
alexcrichton: Weird that was not supposed to sneak in...

src/expr.rs
alexcrichton: Indeed!

PR #194, opened 2017-08-28
Fix `_` as an identifier by accident

alexcrichton:
The proc-macro2 crate currently has a bug where a bare `_` token is classified
as an identifier, but this doesn't happen with the upstream `proc_macro` crate!
In preparation for fixing proc-macro2 this commit removes `syn`'s reliance on
this behavior.


alexcrichton:
This was my "real bug" I was originally trying to hunt down in https://github.com/dtolnay/syn/pull/193

alexcrichton:
Corresponding PR for proc-macro2: https://github.com/alexcrichton/proc-macro2/pull/29

Issue #196, opened 2017-09-07
Allow rust keywords as attribute words/keys


kestred:
I've been writing some derive-based macros, and I found unexpected behavior while parsing an attribute.

Syn will fail to parse a `DeriveInput` if the input's attributes contain a rust keyword.

Some strawman-ish examples:

```rust
#[derive(RepeatStruct)]
#[repeat(for("Hello", "World"))]
struct Input {}

#[derive(SQLExample)]
#[sql(where = "expr IS TRUE")]
struct Thing {}
```

Is this intended behavior?  

If it is undecided, then IMO it'd be great to have it allowed; to me it doesn't seem that the normal rust keywords would ever gain special meaning in core rust attributes.  

If it is a bug, how I could help fix it?

mystor:
Can you test this with the `master` version of `syn`? A lot has changed in how we parse rust code since the last time we made a release, and the new version potentially doesn't have this problem (unfortunately it also has a very different API and hasn't been published yet because it hasn't been polished enough).

kestred:
Just spent some time attempting to update to use the new version--- its turning into quite a lot of work, so that'll wait till the API is more polished.

Instead, hopefully I'll be able to verify with a minimal test case soon.

This should be enough to set it off.
```
#[proc_macro_derive(Any)]
pub fn derive_any(input: TokenStream) -> TokenStream {
    let s = input.to_string();
    let ast = syn::parse_derive_input(&s).unwrap();
    let gen = quote!{};
    gen.parse().unwrap()
}
```

```
#[derive(Any)]
#[any(where = "true"))
struct Test {
}
```

mystor:
I tested this with the master version of syn, and it parses just fine :-).

PR #204, opened 2017-10-24
Add a dump-ast example program for dumping the AST for a single file

mystor:
I wrote this originally to make it easier for me to quickly test stuff like #196, and I figured it might be useful to have in tree. It's just a really simple example program that reads a file, parses it, and debug prints out the result.

I also cleaned up the debug output a bit as it was a little noisy.

### DIFF
--- a/examples/dump-ast.rs
+++ b/examples/dump-ast.rs
@@ -1,0 +1,28 @@
+#![cfg(all(feature = "full", feature = "extra-traits"))]
+extern crate syn;
+
+use std::env;
+use std::fs::File;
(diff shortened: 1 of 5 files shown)


examples/dump-ast.rs
```diff
@@ -0,0 +1,28 @@
+#![cfg(all(feature = "full", feature = "extra-traits"))]
```
alexcrichton: I think this may be best as a `required-features` entry in the manifest, otherwise I think this fails to compile unless the features are set?

PR #206, opened 2017-10-25
Use custom codegen to generate Visitor and Folder source

mystor:
This adds an internal crate, `syn_codegen` which generates the code for `Visitor` and `Folder`, in order to make sure that the code doesn't become out of date over time.

This is probably not ready to merge yet - I'm mostly opening the PR so I can get Travis to take a look at it.


mystor:
(Somewhat related - if anyone has a good idea of how to clean up the codegen logic significantly - it would be fantastic. Ideally I want to keep the output human readable.)

alexcrichton:
Sounds plausible to me!

mystor:
Merging after talking to @dtolnay at rust belt rust.

Issue #212, opened 2017-10-30
Ident needs example code


dtolnay:
Basic things like how to get the string value of an ident, how to build an ident using `format!`.

dtolnay:
Added in #214.

PR #214, opened 2017-10-30
Add examples to Ident

Rantanen:
Added couple of examples to demonstrate the usual needs for the Ident API. Also tweaked the prose a bit to remove the reference to the `syn!` macro. Left the keyword exception in, but it's now in reference to the various `parse_*` functions.

Also added a comment to the `Ident::new` for those people who might skip the examples and are just looking for a way to construct an `Ident`.


Rantanen:
Yeah, I'm somewhat happy how they turned out so happy to contribute!

Also something I forgot to mention in the pull request: Fixes #212, so now you have to close that one manually. :)

PR #217, opened 2017-11-10
Flatten ItemKind into an Item enum

dtolnay:
Side note: this pattern is sort of nice:

```rust
let item: &Item = /* ... */;

match *item {
    Item::Mod(ref item) => {
        do_something_with(&item.ident, &item.content);
    }
    Item::Fn(ref item) => {
        do_something_with(&item.abi, &item.block);
    }
    /* ... */
}
```

Basically pattern matching on an `item` and rebinding the same name `item` just takes your variable and imbues it with those additional fields that exist on the type of item that it happens to be.

### DIFF
--- a/src/gen/fold.rs
+++ b/src/gen/fold.rs
@@ -236,7 +236,7 @@ fn fold_item_foreign_mod(&mut self, i: ItemForeignMod) -> ItemForeignMod { fold_
 # [ cfg ( feature = "full" ) ]
 fn fold_item_impl(&mut self, i: ItemImpl) -> ItemImpl { fold_item_impl(self, i) }
 # [ cfg ( feature = "full" ) ]
-fn fold_item_kind(&mut self, i: ItemKind) -> ItemKind { fold_item_kind(self, i) }
+fn fold_item_mac(&mut self, i: ItemMac) -> ItemMac { fold_item_mac(self, i) }
 # [ cfg ( feature = "full" ) ]
 fn fold_item_mod(&mut self, i: ItemMod) -> ItemMod { fold_item_mod(self, i) }
 # [ cfg ( feature = "full" ) ]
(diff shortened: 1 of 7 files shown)


dtolnay:
@mystor syn_codegen is a godsend, thanks again.

src/gen/fold.rs
```diff
@@ -1544,14 +1544,89 @@ pub fn fold_in_place_kind<V: Folder + ?Sized>(_visitor: &mut V, _i: InPlaceKind)
 }
 # [ cfg ( feature = "full" ) ]
 pub fn fold_item<V: Folder + ?Sized>(_visitor: &mut V, _i: Item) -> Item {
-    Item {
-        attrs: FoldHelper::lift(_i . attrs, |it| { _visitor.fold_attribute(it) }),
-        node: _visitor.fold_item_kind(_i . node),
+    use ::Item::*;
```
gnzlbg: While the pattern might be nice, turning 3 lines of code into ~100 lines of code isn't that nice.

gnzlbg: nvm I see below why this is done, its good :)

mystor: That code is also automatically generated by syn_codegen, so it doesn't really matter how long it is :-).

mystor:
This is pretty nice :-). Can we also do this for Expr, Lit, etc? It would be nice to get rid of more of the `*Kind` types.

dtolnay:
I will have a go at ForeignItemKind, TraitItemKind, ImplItemKind. We need a bit more design work on Lit, the current structure is strange. For Expr I think we can flatten Expr into an enum but we may not be able to expose individual parsers like syn!(ExprBinary) because of how precedence is complicated. I think that is fine, I bet fewer people want to parse a particular kind of Expr than a particular kind of Item.

Issue #224, opened 2017-11-10
Standardize on usage of abbreviations


dtolnay:
For example right now there is Item::Mac but TraitItem::Macro.

What abbreviations do we use? Let's pick one version of each and be consistent. I am okay with choosing independently for each abbreviation, does not have to be all abbreviated or all spelled out.

- - [ ] Mac
  - [x] Macro
- - [ ] Ty
  - [x] Type
- - [ ] Tup
  - [x] Tuple
- - [x] Fn
  - [ ] Function
- - [x] Mod
  - [ ] Module
- - [x] Const
  - [ ] Constant

dtolnay:
Where applicable, I would abbreviate only the things that Rust source code abbreviates. For example Rust uses `fn` and `mod` and `const`, but spells out `type` and `macro`.

PR #241, opened 2017-11-12
Token macros

dtolnay:
I am not dead set on this, but I think it is an improvement. This commit introduces a type macro `Token!` and parser macros `punct!` and `keyword!` used as follows.

<pre>
ItemExternCrate {
    pub attrs: Vec<Attribute>,
    pub vis: Visibility,
    pub extern_token: <b>Token![extern]</b>,
    pub crate_token: <b>Token![crate]</b>,
    pub ident: Ident,
    pub rename: Option<(<b>Token![as]</b>, Ident)>,
    pub semi_token: <b>Token![;]</b>,
}

impl_synom!(ItemExternCrate "extern crate item" do_parse!(
    attrs: many0!(call!(Attribute::parse_outer)) >>
    vis: syn!(Visibility) >>
    extern_: <b>keyword!(extern)</b> >>
    crate_: <b>keyword!(crate)</b> >>
    ident: syn!(Ident) >>
    rename: option!(tuple!(<b>keyword!(as)</b>, syn!(Ident))) >>
    semi: <b>punct!(;)</b> >>
    (ItemExternCrate {
        attrs: attrs,
        vis: vis,
        extern_token: extern_,
        crate_token: crate_,
        ident: ident,
        rename: rename,
        semi_token: semi,
    })
));
</pre>

There are a few advantages.

- I learn this scheme once and then I can be productive. In the previous scheme I frequently have to look up the names of RArrow, Colon2, Star, etc.

- No longer misleading for punctuation that has overloaded meanings. In the previous scheme for example TyRptr, which represents `&T` and `&mut T`, contained a `tokens::And`. But in this context the `&` symbol means "reference", not "and".

- No longer need to glob import `synom::tokens::*` which means token names no longer conflict with prelude types or syn types. As part of this commit, I have renamed `tokens::Box_`, `tokens::Default_`, and `tokens::Fn_` to remove the underscores. I kept `tokens::Self_` because that one would be a reserved word without underscore. This also frees up the name `Type` for #224.

### DIFF
--- a/src/attr.rs
+++ b/src/attr.rs
@@ -9,7 +9,7 @@ ast_struct! {
     /// Doc-comments are promoted to attributes that have `is_sugared_doc` = true
     pub struct Attribute {
         pub style: AttrStyle,
-        pub pound_token: tokens::Pound,
+        pub pound_token: Token![#],
         pub bracket_token: tokens::Bracket,
 
         /// The path of the attribute.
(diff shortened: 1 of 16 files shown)


mystor:
I like this a lot. I'd say that it's definitely a usability improvement over the old system.

alexcrichton:
👍

Issue #232, opened 2017-11-11
Change round trip tests to use libsyntax instead of syntex


dtolnay:
Keeping Syntex up to date is more work than testing with nightly Rust and unstable libsyntax.

PR #248, opened 2017-11-13
Test with libsyntax instead of syntex

dtolnay:
- [ ] Make it compile with the latest nightly
- [ ] Adjust CI script to only run these tests on nightly

#232


mystor:
I'll drive this to the finish line sometime today :-).

Issue #227, opened 2017-11-11
Update AST to hold const generics


dtolnay:
The Generics struct now needs to hold `lifetimes`, `ty_params`, and **`const_params`**.

https://github.com/rust-lang/rust/issues/44580

```rust
struct RectangularArray<T, const WIDTH: usize, const HEIGHT: usize> {
    array: [[T; WIDTH]; HEIGHT],
}
```

eddyb:
That's not how we're doing it in the compiler, see https://github.com/rust-lang/rust/pull/45930.

dtolnay:
Implemented in #261.

PR #261, opened 2017-12-04
Generic arg refactor

### LINKED ISSUE #259: Parse PolyTraitRef with lifetime before type
This syntax is allowed since Rust 1.19.0.

```rust
extern crate syn;

fn main() {
    syn::parse_str::<syn::DeriveInput>(
        "struct S<'a>(Box<'a + Iterator<Item = usize>>);"
    ).unwrap();
}
```

mystor:
This is built on top of #260.

This fixes #259.

### DIFF
--- a/src/expr.rs
+++ b/src/expr.rs
@@ -202,9 +202,14 @@ ast_enum_of_structs! {
             pub or2_token: Token![|],
         }),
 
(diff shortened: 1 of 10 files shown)


mystor:
I added some commits to also handle parsing const parameters, which should hopefully close #227.

Issue #120, opened 2017-04-21
Can't parse type parameters containing trailing +


Arnavion:
```rust
let raw = r#"
    fn main() {
        let x: Box<Debug+>;
    }
"#;

let actual = parse_item(raw).unwrap(); // Fails
```

Found in [`tests/rust/src/test/ui/token/trailing-plus-in-bounds.rs`](https://github.com/rust-lang/rust/blob/ddc5d7bd4b9ea3e8a8ccf82cb443e950be311d61/src/test/ui/token/trailing-plus-in-bounds.rs#L16) which should pass without the trailing `FAIL`

dtolnay:
Wow I had no idea this was allowed!

mystor:
Shouldn't be hard so putting it in the 0.12 milestone.

dtolnay:
Fixed in #265.

eddyb:
This was mistakenly fixed only as a special case of a general argument, but [it's valid syntax for any `dyn`/`impl` (including old `dyn`-less trait objects)](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=4a62af0321276bd9ca9af4911dfbf53d).

Filing a new issue now.

PR #265, opened 2017-12-05
Generic assoc types

### LINKED ISSUE #226: Update AST to hold generic associated types
TraitItemType and ImplItemType will need to hold generic parameters, not just an ident for the associated type.

https://github.com/rust-lang/rust/issues/44265

```rust
trait StreamingIterator {
    type Item<'a>;
    fn next<'a>(&'a mut self) -> Option<Self::Item<'a>>;
}
```

mystor:
Fixes #226


mystor:
I threw in a fix for #120 as well, because I'm terrible at keeping patches separate ^.^

I didn't add the where clause to the impl type, because I couldn't get any code to compile which used one on play.rust-lang.org.

dtolnay:
Elsewhere I think we put the where-clause [in the Generics](https://github.com/dtolnay/syn/blob/4fee25e6c77bafd2820a1c2290386dd0b96d81de/src/generics.rs#L12).

mystor:
Whoops - should be fixed now

PR #266, opened 2017-12-05
Add a feature to synom for verbose parsing logging

mystor:
I'm not sure if we want this to be in the tree, but I've written code which does something like this to debug why and where a parse is failing multiple times.

I intentionally don't expose this as a feature on syn, as I don't want --all-features on syn to pick it up ('cause we use that for tests), but this unfortunately makes it a bit hard to use this feature when testing. I usually edit synom's Cargo.toml when I want to turn it on.

### DIFF
--- a/synom/Cargo.toml
+++ b/synom/Cargo.toml
@@ -22,6 +22,10 @@ default = []
 clone-impls = []
 extra-traits = []
 
+# NOTE: This feature is very verbose - it will cause every call!() expression to
+# be logged to stderr. It exists mostly as a debugging tool.
+verbose-trace = []
+
 [dev-dependencies.syn]
 version = "0.11"
 path = ".."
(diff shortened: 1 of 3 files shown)


mystor:
Here's an example snippet of output for parsing a 'a as part of a trait object.

```
 -> ty_poly_trait_ref @ Cursor("\'a > { }")
 -> Delimited::parse_terminated_nonempty @ Cursor("\'a > { }")
 -> <::tokens::Question as ::Synom>::parse @ Cursor("\'a > { }")
ERR <::tokens::Question as ::Synom>::parse @ Cursor("\'a > { }")
 -> <Lifetime as ::Synom>::parse @ Cursor("\'a > { }")
OK  <Lifetime as ::Synom>::parse @ Cursor("> { }")
OK  Delimited::parse_terminated_nonempty @ Cursor("> { }")
OK  ty_poly_trait_ref @ Cursor("> { }")
```

PR #274, opened 2017-12-19
Support folding/visiting over Ident and Span tokens

mystor:


mystor:
This is handy for doing things like manipulating / investigating spans. I also fold over `Ident`s in synstructure on syn 0.11.

PR #279, opened 2017-12-22
Add From<Span> for all tokens

### LINKED ISSUE #278: provide some helpers for span.
- `span.new_ident`

 - `span.as_token()`
Creates `tokens::Token` from span. In my own experience, it's **really** useful for handling span.
Span created by `Default::default()` (which is `call_site`) isn't useful for error reporting.

e.g. span of some special tokens like `impl` in `impl Trait` is used for type mismatch.

-----

https://github.com/kdy1/rust-pmutil/blob/3827505add14547b56c027cdd124a73e52527c4e/src/synom_ext.rs
```rust
pub trait FromSpan {
    fn from_span(span: Span) -> Self;
}
pub trait SpanExt: Copy {
    fn new_ident<S>(self, s: S) -> syn::Ident
    where
        S: AsRef<str>;

    /// Creates `Token` from `self`.
    fn as_token<Token>(self) -> Token
    where
        Token: FromSpan;
}
```
(maybe need to add `from_spans`?)

-----

I'll send a pr for this.

kdy1:
Fixes #278.


kdy1:
I'm happy with Ident::new(), but I'm not sure if Else(span) is better.
Currently, span from procedural macros work in strange way.
In my experience, which **span is used** is important than which token is I'm currently creating.

For example, rustc fails to resolve module if span from function body is used in `::crate_name` (first part) of return type.

dtolnay:
@mystor could you take a look and recommend a path forward? Thanks!

kdy1:
IMHO, it's similar to `From<Span>` and `Into<token::XX>`. Implementing `From<Span>` for tokens and using `span.into()` would be better than `span.as_token()`.

While creating pr, I completely forgot why I chose `span.as_token()` instead of `span.into()`.

dtolnay:
I would be on board with providing `From<Span>` for all the token types. Good idea! Could you rebase the PR and change it to `From` impls?

kdy1:
I'll do it soon.

kdy1:
I did it.
(I made a mistake because ide didn't show it...)

dtolnay:
I published this in syn 0.12.5.

PR #314, opened 2018-01-10
Add run_parser* functions

### LINKED ISSUE #313: Improve the situation for parsing using specialized parser methods.
e.g. we don't have a mechanism for easily parsing a Punctuated (e.g. `Punctuated::parse_terminated`) using `syn::parse{, 2, _str}`, It'd be nice if there was some helper mechanism for performing that kind of function based parse as well.

Perhaps `syn::run_parser{, 2, _str}` which takes a `fn (c: Cursor) -> PResult<T>` and returns a `Result<T, ParseError>`? It'd be trivial to implement the existing `syn::parse` on top of it.

cc @dtolnay

mystor:
I don't really like these function names - I'd be very open to coming up to better names for the functions.

It also might be a good idea to merge these all into a single `run_parser` which takes an `Into<proc_macro2::TokenStream>` given that the superfish won't be used often with these types, as the type of the function argument usually cannot be written.

I also changed `parse2` to call `run_parser2` under the hood, as I figured that keeping the logic in one place was a good idea.

Fixes #313

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -594,25 +594,12 @@ pub fn parse2<T>(tokens: proc_macro2::TokenStream) -> Result<T, ParseError>
 where
     T: Synom,
 {
-    let buf = TokenBuffer::new2(tokens);
-    let result = T::parse(buf.begin());
-    let err = match result {
-        Ok((t, rest)) => {
-            if rest.eof() {
-                return Ok(t);
-            } else if rest == buf.begin() {
-                // parsed nothing
-                ParseError::new("failed to parse anything")
-            } else {
-                ParseError::new("failed to parse all tokens")
-            }
+    run_parser2(tokens, T::parse).map_err(|err| {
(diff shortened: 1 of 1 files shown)


dtolnay:
Do you think this would work well as an extension trait?

```rust
impl<F, R> SynomExt for F where F: FnOnce(Cursor) -> PResult<R> {
    fn parse;
    fn parse2;
    fn parse_str;
}

fn main() {
    let exprs = Punctuated::<Expr, Token![,]>::parse_terminated;

    for e in exprs.parse_str("a, b, 3").unwrap() {
        println!("{:?}", e);
    }
}
```

mystor:
I think I'd like it better if the trait was called `Parser` rather than `SynomExt`, because it's a trait for describing something which can parse using syn's parsing system. The `SynomExt` name feels gross, but just changing it to `Parser` doesn't IMO?

Issue #315, opened 2018-01-11
Can't `parse_quote!` an Attribute


mystor:
It's unfortunate that it's not possible to `parse_quote!` an attribute. It'd be good if we could use `parse_quote!` to parse one. I think this would involve implementing `Synom` for `Attribute`, and parsing outer attributes in it.

(It probably makes much more sense to parse outer attributes there than inner ones).

dtolnay:
Why not parse both? `parse_quote!(#[...])` gives you an attribute with style AttrStyle::Outer and `parse_quote!(#![...])` gives you an attribute with style AttrStyle::Inner.

mystor:
I'd be fine with that too. The worry is that `syn!(Attribute)` would Do The Wrong Thing in many cases, and I'm worried about encouraging that.

dtolnay:
Yeah. We could have `parse_quote!` use a different trait that is implemented for all `T: Synom` but also some other types.

PR #318, opened 2018-01-11
Provide a trait to invoke parser functions

### LINKED ISSUE #313: Improve the situation for parsing using specialized parser methods.

dtolnay:
As discussed in #314. Closes #314, fixes #313.

### DIFF
--- a/src/synom.rs
+++ b/src/synom.rs
@@ -8,14 +8,115 @@
 
 //! Parsing interface for parsing a token stream into a syntax tree node.
 //!
-//! All syntax tree nodes that can be parsed from a token stream need to
-//! implement the [`Synom`] trait. This trait is usually implemented using the
-//! [`nom`]-style parser combinator macros provided by Syn, but may also be
-//! implemented without macros be using the low-level [`Cursor`] API directly.
+//! Parsing in Syn is built on parser functions that take in a [`Cursor`] and
+//! produce a [`PResult<T>`] where `T` is some syntax tree node. `Cursor` is a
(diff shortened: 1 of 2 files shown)


src/synom.rs
```diff
@@ -8,14 +8,115 @@
...
+//! [`ParseError`]: struct.ParseError.html
+//!
+//! This `Cursor`- and `PResult`-based interface is convenient for parser
+//! combinators and parser implementations, but not necessarily when you just
+//! have some tokens that you want to parse. For that we expose the following
+//! two entry points.
+//!
+//! ## The `syn::parse*` functions
+//!
+//! The [`syn::parse`], [`syn::parse2`], and [`syn::parse_str`] functions serve
```
mystor: Mention parse_quote! here?

dtolnay:
I am going to work on #315 and then document how `parse_quote!` fits in. I think we want `parse_quote!` to be able to parse both Attribute and Punctuated.

src/synom.rs
dtolnay: Added a note in #319.

PR #319, opened 2018-01-11
Some special cases for parse_quote

### LINKED ISSUE #315: Can't `parse_quote!` an Attribute

dtolnay:
Fixes #315.

### DIFF
--- a/src/parse_quote.rs
+++ b/src/parse_quote.rs
@@ -1,0 +1,156 @@
+/// Quasi-quotation macro that accepts input like the [`quote!`] macro but uses
+/// type inference to figure out a return type for those tokens.
+///
+/// [`quote!`]: https://docs.rs/quote/0.4/quote/index.html
+///
+/// The return type can be any syntax tree node that implements the [`Synom`]
+/// trait.
+///
+/// [`Synom`]: synom/trait.Synom.html
+///
+/// ```
+/// #[macro_use]
+/// extern crate syn;
+///
+/// #[macro_use]
+/// extern crate quote;
+///
+/// use syn::Stmt;
+///
+/// fn main() {
+///     let name = quote!(v);
+///     let ty = quote!(u8);
+///
+///     let stmt: Stmt = parse_quote! {
+///         let #name: #ty = Default::default();
+///     };
+///
+///     println!("{:#?}", stmt);
+/// }
+/// ```
+///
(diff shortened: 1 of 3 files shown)


src/parse_quote.rs
```diff
@@ -0,0 +1,156 @@
+/// Quasi-quotation macro that accepts input like the [`quote!`] macro but uses
+/// type inference to figure out a return type for those tokens.
+///
+/// [`quote!`]: https://docs.rs/quote/0.4/quote/index.html
+///
+/// The return type can be any syntax tree node that implements the [`Synom`]
```
mystor: Perhaps mention the special cases here under a subheading?

src/parse_quote.rs
```diff
@@ -0,0 +1,156 @@
...
+////////////////////////////////////////////////////////////////////////////////
+// Can parse any type that implements Synom.
+
+use quote::Tokens;
+use synom::{Synom, Parser, PResult};
+use buffer::Cursor;
+
+// Not public API.
+#[doc(hidden)]
+pub fn parse<T: ParseQuote>(tokens: Tokens) -> T {
```
mystor: Why not put this on ParseQuote directly instead of having a freestanding function?

src/parse_quote.rs
dtolnay: Yep I have a "Special cases" heading below. I would prefer to keep them below the examples because most people are not going to keep straight which types do and don't implement Synom. They will just see the examples, understand the concept, and then try it with Attribute and have it silently do the thing they wanted.

src/parse_quote.rs
dtolnay: I think I would move this logic into a method on ParseQuote if that were able to replace the `(Cursor) -> PResult<Self>` method with a `(Tokens) -> Self` method, but I don't think that would work because all the impls want to provide the `(Cursor) -> PResult<Self>` signature and so it would need to be a third method. If it is a separate method either way rather than replacing a method from ParseQuote, then I would keep it here by analogy with `syn::parse`.

Anyway, it makes sense to me to conceptualize ParseQuote as "just Synom but with some extra implementations."

PR #327, opened 2018-01-15
Add helpful accessors to Generics

sgrif:
NOTE: This PR is not yet ready to be merged. This adds only two of the
six needed accessors, and has no tests. I wanted to make sure that
this was a feature that would be accepted before spending the time to
finish this.

Within Diesel we have a bunch of places that do things like
`generics.params.iter()`. It'd be very helpful for syn to provide these
iterators for us.

This PR as written only contains the iterator for `type_params` (holy
crap I hate that name can we please rename `TypeParam` to
`TypeParameter`). The other iterators will be mechanically
identical.

If you're good with this feature I will implement the remaining
iterators and add tests.

### DIFF
--- a/src/generics.rs
+++ b/src/generics.rs
@@ -7,7 +7,7 @@
 // except according to those terms.
 
 use super::*;
-use punctuated::Punctuated;
+use punctuated::{Iter, IterMut, Punctuated};
 
 ast_struct! {
     /// Lifetimes and type parameters attached to a declaration of a function,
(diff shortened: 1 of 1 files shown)


dtolnay:
I am on board with adding these. Thanks!

I would be comfortable accepting these without tests.

src/generics.rs
```diff
@@ -77,6 +77,61 @@ ast_enum_of_structs! {
...
+/// A mutable iterator over the type parameters of `Generics`.
+///
+/// Returned by `Generics::type_params_mut`
+pub struct TypeParamsMut<'a>(IterMut<'a, GenericParam, Token![,]>);
+
+impl<'a> Iterator for TypeParamsMut<'a> {
+    type Item = &'a mut TypeParam;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        // FIXME: Remove this when ? on Option is stable
```
mystor: Syn supports rust versions back to 0.15, so unless we increase our minimum rust version, we won't be able to switch to using ? on Option when it becomes stable.

sgrif: In that case #324 needs to be reverted, since it doesn't compile on 1.15. It's probably worth adding 1.15 to CI if the intent is to continue to support it in all future versions. (I personally don't think it's worth going much farther than the rust-lang-nursery policy)

mystor: It is being tested in CI (https://github.com/dtolnay/syn/blob/03a092060471ed8bddf492c537d980a57b8f511d/.travis.yml#L18-L19), but travis was choking & I foolishly merged without letting the CI finish.

@dtolnay is the person who has picked our minimum version. I think that our version minimum is 0.15 to match serde's version minimum? Can't remember exactly.

I'll fix up the code to build locally.

sgrif:
I've added the remaining accessors

src/generics.rs
sgrif: You mean 1.15 not 0.15, right?

mystor: yes, 1.15 not 0.15... oops.

mystor: Fixed in 5680dcc26cbfce6b8590e898df884f47d03c8bed

PR #363, opened 2018-02-23
Add a feature for linking to proc_macro.

staktrace:
This feature is enabled by default for backwards compatibility, but
it can be disabled in order to break the runtime dependency on the
dynamic library libproc_macro in the rustc toolchain.

This PR assumes that there is a version 0.5.0 or higher of `quote` on crates.io that includes dtolnay/quote#62. That is not yet the case (see https://github.com/dtolnay/quote/pull/62#issuecomment-368087621)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,7 +15,7 @@ path = "examples/dump-syntax/main.rs"
 required-features = ["full", "parsing", "extra-traits"]
 
 [features]
-default = ["derive", "parsing", "printing", "clone-impls"]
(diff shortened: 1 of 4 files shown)


staktrace:
I wasn't sure what travis.yml changes you wanted here, since it looks like it builds in some subset of possible configurations but only tests the configuration with all the features.

dtolnay:
Thank you! I filed #369 to reopen and merge this PR during the next round of breaking changes.

PR #367, opened 2018-02-28
Document error handling

federicomenaquintero:


federicomenaquintero:
So, I looked to see what Serde does, as I've never used it and don't know what its error messages look like.  If I'm reading this correctly, `serde_derive_internals/src/attr.rs` -> `impl Container` -> `fn from_ast()` - that one checks for the `rename` you mentioned.  If a parse error occurs it just aggregates a string for the error into the `Ctxt`.

Later, `serde_derive/src/ser.rs` in `expand_derive_serialize()` calls `ctx.check()`, which is what produces a `Result<..., with_that_chain_of_error_strings>`.  

Finally, `serde_derive` has this in one of its entry points:

```rust
#[proc_macro_derive(Serialize, attributes(serde))]
pub fn derive_serialize(input: TokenStream) -> TokenStream {
    let input: DeriveInput = syn::parse(input).unwrap();
    match ser::expand_derive_serialize(&input) {
        Ok(expanded) => expanded.into(),
        Err(msg) => panic!(msg),
    }
}
```

It's just doing a `panic!` with the resulting string.  I have two questions :smile: 

1. I guess Serde should be doing `compile_error!` there, both instead of `unwrap()` and the `panic!`?
1. Strategies for handling/presenting errors at various levels.  Okay, this is like a million questions.

From the way the gnome-class code is turning out, it looks like it will have places that generate errors at various levels.  I suppose Serde is in the same situation.

* Errors during parsing, like your `serde(rename())` example.  For these it is convenient to just bubble up a `Result` until you extract the error description from it at the toplevel, and return `quote! { compile_error!(#descr); }`.  I *think* in all cases the compiler will present such an error as spanning the whole invocation of the macro, which is suboptimal?  E.g. I'd like an error to come from Syn's innards to say something like, "expected 'fn' but found 'blah' instead".  I haven't looked closely enough at the synom sources to see if this is possible.

* Errors detected during an "analysis" stage.  Implementations of proc macros will more or less create their own AST out of the parsed stuff, and then validate it / transform it in other ways.  The `lazy_static` example shows that to present good, spanned error messages for those, one should call `problematic_token.span().unstable().warning("blah blah").emit()` (or `error()`), and then quote that token into the final `TokenStream` as if everything were okay.  Then the compiler will present the warnings/errors in the right place.  Is this correct?

* Related to the above - if the analysis stage just cannot proceed after detecting an error, i.e. if it can't generate a `TokenStream` "as if everything were okay"... I suppose one can bail out, and just emit the usual `quote! { #problematic_token_with_the_span_error_already_called }`?

I'm happy to write these docs; I'd just like confirmation that I'm understanding things correctly :smile: I need to write more examples for myself and see what happens!

PR #371, opened 2018-03-09
deps: update walkdir to 2

ignatenkobrain:


ignatenkobrain:
Yeah, sorry. Will fix it asap.

ignatenkobrain:
@dtolnay should be good now. Sorry for delay.

tests/test_precedence.rs
```diff
@@ -97,7 +97,7 @@ fn test_rustc_precedence() {
     let failed = AtomicUsize::new(0);
 
     WalkDir::new("tests/rust")
-        .sort_by(|a, b| a.cmp(b))
+        .sort_by(|a, b| a.file_name().cmp(b.file_name()))
```
dtolnay: I believe it would make more sense to sort by the full path not the filename. Lots of different paths may have the same filename (like lib.rs) and we don't want those in the output in random order.

Issue #304, opened 2018-01-08
Syntax tree [src] links all point to macros.rs


dtolnay:
For example the [src] link from [`syn::Abi`](https://docs.rs/syn/0.12.0/syn/struct.Abi.html) goes [here](https://docs.rs/syn/0.12.0/src/syn/macros.rs.html#47-207) rather than to where you probably want to go in ty.rs.

From experimenting a bit, it seems that the link goes wherever the `pub` token originates from. In this case macros.rs is discarding the `pub` token from the call site and coming up with its on `pub` token, which is why the links go where they do.

```rust
(
    $(#[$attr:meta])*
    pub struct $name:ident $($rest:tt)*
) => {
    $(#[$attr])*
    #[cfg_attr(feature = "extra-traits", derive(Debug, Eq, PartialEq, Hash))]
    #[cfg_attr(feature = "clone-impls", derive(Clone))]
    pub struct $name $($rest)*
};
```

Ideally we could preserve the original `pub` token by matching `$vis:ident struct $name:ident` and emitting `$vis struct $name`, but that triggers a macro_rules ambiguity prior to Rust 1.20.0.

```
error: local ambiguity: multiple parsing options: built-in NTs ident ('vis') or 1 other option.
   --> src/macros.rs:91:19
    |
91  |                   $(#[$variant_attr])*
    |                     ^
```

We may need to go with either a TT-muncher or some dumb syntax that resolves the ambiguity and lets us keep the `pub` token on 1.15.0, like:

```rust
ast_struct! {
    /// ...
    @pub struct Abi {
```

dtolnay:
Filed https://github.com/rust-lang/rust/issues/47817 to consider fixing this in rustdoc.

dtolnay:
Fixed in #376 and 7f2dce7ede0e0d8c973a88f76e61544393be2d3a + e3a6066517dab34ecd6715bf4f9c87b624cf9908.

PR #376, opened 2018-03-19
preserve spans in macros to generate better rustdoc [src] links

mystor:
Make rustdoc generate the correct `[src]` links by perserving spans in our macros.

`__check` makes sure you don't write the wrong keywords.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -23,6 +23,10 @@ fn main() {
         println!("cargo:rustc-cfg=syn_can_use_associated_constants");
     }
 
+    if compiler.minor >= 20 {
+        println!("cargo:rustc-cfg=syn_can_match_ident_after_attrs");
+    }
+
     // Macro modularization allows re-exporting the `quote!` macro in 1.30+.
     if compiler.minor >= 30 {
         println!("cargo:rustc-cfg=syn_can_call_macro_by_path");
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -286,6 +286,7 @@ extern crate unicode_xid;
 #[cfg(feature = "printing")]
 extern crate quote;
 
+#[cfg(any(feature = "full", feature = "derive"))]
 #[macro_use]
 mod macros;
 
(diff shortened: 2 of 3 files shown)


dtolnay:
Is this for #304? I think `$(#[$attr:meta])* $pub:ident` is an ambiguity prior to Rust 1.20.

mystor:
Yeah, this is for #304. Shucks re: the ambiguity thing :cry: 

I'll have to fiddle around a bit more to see if I can work around that while still preserving the spans...

mystor:
@dtolnay With 1.0 bumping the min rustc version this may be worth resurrecting :-)

mystor:
I updated it, added some fallback logic. I had to do some additional minor changes to ensure that as many items as possible get correctly spanned, because if rustc sees that either the first or last span in the item is attributed to a macro, it attributes the entire span to the macro.

Unfortunately, I was unable to get the spans generated the outer enum from `ast_enum_of_structs!` to be correct (though I was able to get correct spans for the inner generated structs). I added a comment to the code to explain why (copied below): 

```
// Unfortunately, at this time, we can't make the generated enum here have the
// correct span. The way that the span for the overall enum decl is calculated
// is using `lo.to(prev_span)` [1].
//
// If the beginning and ending spans of the enum item don't have the same macro
// context, fallback code is run [2]. This code chooses the span within the
// macro context to avoid firing spurious diagnostics [3].
//
// This means `[src]` links in rustdoc will point to the macro, instead of the
// actual declaration, if either the first or last token of our declaration has
// a span from the macro. With `macro_rules!` there is no way to preserve the
// span of a `{}` block while changing its contents, so we're forced to have the
// span of our final token point to our macro.
//
// [1]: https://github.com/rust-lang/rust/blob/9a90d03ad171856dc016c2dcc19292ec49a8a26f/src/libsyntax/parse/parser.rs#L7377
// [2]: https://github.com/rust-lang/rust/blob/9a90d03ad171856dc016c2dcc19292ec49a8a26f/src/libsyntax_pos/lib.rs#L467-L478
// [3]: https://github.com/rust-lang/rust/pull/47942
```

PR #405, opened 2018-04-20
Implement AsRef<Path> for Ident

clarcharr:
This will make it a bit easier to create paths to source files from `Ident`s.

Example use case: traversing an `ItemUse` statement will give you a `Vec<Ident>` path to an item. To get the path to the module file, you can simply do:

```
let mut buf = PathBuf::from(SRC_DIR);
buf.extend(path);
buf.set_extension("rs");
```

whereas before you had to do:

```
let mut buf = PathBuf::from(SRC_DIR);
buf.extend(path.into_iter().map(AsRef::as_ref));
buf.set_extension("rs");
```

To support this use case, `self` and `super` idents will give the paths `.` and `..`.

### DIFF
--- a/src/ident.rs
+++ b/src/ident.rs
@@ -10,6 +10,7 @@ use std::borrow::Cow;
 use std::cmp::Ordering;
 use std::fmt::{self, Display};
 use std::hash::{Hash, Hasher};
+use std::path::Path;
 
 use proc_macro2::Term;
 use unicode_xid::UnicodeXID;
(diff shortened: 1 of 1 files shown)


clarcharr:
Did a quick scrape of crates.io and I couldn't find any crates that would do `File::open` based upon an `Ident`, so ¯\_(ツ)_/¯. I also couldn't find any crates that made use of `parse_file` based upon module imports, as most people opted to use `walkdir` or only looked at `lib.rs` or some other hard-coded file.

In case anyone else finds this interesting, I found the crates to look through with `fd` and this `jq` command on the `crates.io-index` repo:

```
jq '[.deps[] | select(.name == "syn") | select(.features | contains(["full"]))] | length > 0'
```

Normally I'd say this is isn't a big maintenance burden but the fact that `as_ref` is now ambiguous is probably enough of a resaon to not merge it. Pretty much up to you.

PR #424, opened 2018-05-12
Introduce `Ident::parse_any` for parsing identifiers with rust keywords

MSleepyPanda:


dtolnay:
For your use case is it better to treat `_` as a raw ident or not? Why?

MSleepyPanda:
My usecase is a GraphQL parser. The spec [allows single `_` as names](http://facebook.github.io/graphql/draft/#sec-Names).

dtolnay:
Seems reasonable, I am on board with accepting `_` here.

I would prefer to avoid introducing a `RawIdent` type because this is just an alternative way to parse an `Ident`. Maybe you could do what we have for [`Field`](https://docs.rs/syn/0.13/syn/struct.Field.html) where we expose `call!(Field::parse_named)` and `call!(Field::parse_unnamed)` for struct fields and tuple struct fields respectively. Here we would have `syn!(Ident)` for the default behavior and `call!(Ident::parse_any)` to accept any ident including keywords.

MSleepyPanda:
Seems more than reasonable! Updated the PR

dtolnay:
I released 0.13.10 containing this function.

PR #426, opened 2018-05-16
Update to the next version of proc-macro2

alexcrichton:
~~Depends on dtolnay/quote#73~~
~~Depends on alexcrichton/proc-macro2#90~~
~~Depends on a new nightly~~


src/synom.rs
```diff
@@ -213,6 +213,16 @@ impl Synom for proc_macro2::TokenStream {
     }
 }
 
+impl Synom for proc_macro2::Ident {
+    fn parse(input: Cursor) -> PResult<Self> {
+        input.term().ok_or_else(|| ParseError::new("not an ident"))
```
dtolnay: I would like this to preserve the behavior of 0.13's impl Synom for Ident. Separately we will need some way to do https://docs.rs/syn/0.13/syn/struct.Ident.html#method.parse_any.

src/synom.rs
```diff
@@ -213,6 +213,16 @@ impl Synom for proc_macro2::TokenStream {
...
+    }
+
+    fn description() -> Option<&'static str> {
+        Some("arbitrary token stream")
```
dtolnay: This should be Some("identifier").

alexcrichton:
Ok it looks like the last failure is 1.15.1 but that's just because it doesn't understand `[patch]` and should be fixable with a new release of `quote`

codegen/src/main.rs
```diff
@@ -106,7 +107,7 @@ fn load_file<P: AsRef<Path>>(name: P, features: &Tokens, lookup: &mut Lookup) ->
                 // We don't want to try to load the generated rust files and
                 // parse them, so we ignore them here.
                 for name in IGNORED_MODS {
-                    if item.ident == name {
+                    if item.ident.to_string() == *name {
```
dtolnay: I opened https://github.com/alexcrichton/proc-macro2/pull/92 to avoid these changes.

src/lifetime.rs
```diff
@@ -140,11 +149,13 @@ pub mod parsing {
...
     use super::*;
-    use quote::{ToTokens, Tokens};
+    use quote::{ToTokens, TokenStreamExt};
+    use proc_macro2::{TokenStream, Punct, Spacing};
 
     impl ToTokens for Lifetime {
-        fn to_tokens(&self, tokens: &mut Tokens) {
-            self.term.to_tokens(tokens);
+        fn to_tokens(&self, tokens: &mut TokenStream) {
+            tokens.append(Punct::new('\'', Spacing::Joint));
```
dtolnay: This token needs to be stored in the Lifetime struct so that we preserve its Span correctly.

src/ident.rs
```diff
@@ -1,306 +0,0 @@
...
-/// if ident_str.len() > 60 {
-///     println!("Very long identifier: {}", ident_str)
-/// }
-///
-/// // Create a String from the ident.
-/// let ident_string = ident.to_string();
-/// give_away(ident_string);
-///
-/// fn give_away(s: String) { /* ... */ }
-/// ```
```
dtolnay: There is valuable explanation and examples being removed here. Compare:

- https://docs.rs/syn/0.13/syn/struct.Ident.html vs
- https://docs.rs/proc-macro2/0.4.1/proc_macro2/struct.Ident.html

Any thoughts on how to provide Syn-specific Ident documentation?

src/gen/fold.rs
```diff
@@ -191,8 +192,6 @@ fn fold_generic_method_argument(&mut self, i: GenericMethodArgument) -> GenericM
 fn fold_generic_param(&mut self, i: GenericParam) -> GenericParam { fold_generic_param(self, i) }
 # [ cfg ( any ( feature = "full" , feature = "derive" ) ) ]
 fn fold_generics(&mut self, i: Generics) -> Generics { fold_generics(self, i) }
-
-fn fold_ident(&mut self, i: Ident) -> Ident { fold_ident(self, i) }
```
dtolnay: I would like to keep fold_ident.

dtolnay:
I can't comment on it directly but please remove the tests/rust submodule added in this PR. IIRC we used to have it but it slows down Cargo git dependencies on Syn.

src/gen/fold.rs
alexcrichton: Oops didn't even realize that went away

src/ident.rs
alexcrichton: Hm I wonder if we could just wholesale move the documentation to `proc_macro2::Ident`?

src/ident.rs
```diff
@@ -1,306 +0,0 @@
...
-        fn description() -> Option<&'static str> {
-            Some("identifier")
-        }
-    }
-    
-    impl Ident {
-        /// Parses any identifier
-        /// 
-        /// This is useful when parsing a DSL which allows Rust keywords as identifiers.
-        pub fn parse_any(input: Cursor) -> PResult<Self> {
```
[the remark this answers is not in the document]
alexcrichton: Perhaps a new macro parser in synom, like `any_ident!()`?

dtolnay:
I think we are mostly waiting for @Eijebong to wrap up work on https://github.com/servo/servo/issues/20497 so that the timing can be maximally bad.

After that, docs need updating, we need some design work around how to handle what used to be `Ident::parse_any`, and I would like to get serde_derive updated before publishing so that we understand practical implications of the changes. If you have bandwidth to help, the best thing would be to make a PR for Serde with some `[patch.crates-io]` dependencies on this branch.

src/ident.rs
dtolnay: Sounds good, I moved this to proc-macro2 in https://github.com/alexcrichton/proc-macro2/commit/8b71daca2dd228053ac4df17a356f667332158fd.

src/ident.rs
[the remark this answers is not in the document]
dtolnay: I implemented this as an extension trait in fe68c90ec8af31a7bbbf554405c1b7e34ae4b0ed.

dtolnay:
Serde update: https://github.com/serde-rs/serde/pull/1273

PR #450, opened 2018-07-20
wip: support for async functions and closures

ubnt-intrepid:
The async/await syntax has already available on the nightly compiler and the authors of proc-macro can experiment by supporting `syn` for parsing its syntax.

The full implementation has not completed yet, but I open this PR to clarify that the work is now ongoing.

@dtolnay if you have some implementation plan on this syntax, could you let me know it?

### Tasks
- [ ] async functions (fixes #395)
- [ ] async closures (fixes #396)

Issue #395, opened 2018-04-06
Parse async functions


dtolnay:
https://github.com/withoutboats/rfcs/blob/async-await/text/0000-async_await.md#async-functions

```rust
async fn function(argument: &str) -> usize {
     // ...
}
```

steveklabnik:
Getting this would be awesome 👍

dtolnay:
@ubnt-intrepid or @steveklabnik, to get this working in the near term would one of you be willing to send a PR that parses async functions as a `syn::Item::Verbatim`? We haven't made use of this variant so far but it is designed for short-term handling of unstable syntax without a breaking change.

steveklabnik:
I might be willing to; I'm able to work around this restriction for now so it's not *super* pressing for me, but if I can find the time I'll give it a shot :)

(Or, if it's easy for you to point out the general gist of how it's done; a lot of the effort would be in figuring out how `syn` works at all, since I haven't delved into the code yet)

dtolnay:
Oh, here is a good example of how to do this: you have to swap red and green in your head but https://github.com/dtolnay/syn/commit/b21b03437310c1c820625e68de68a4fbfdf4edf0 shows how we used to parse the unstable syntax `impl Trait for .. {}` as verbatim tokens.

ubnt-intrepid:
@dtolnay and @steveklabnik Sorry for late reply.
I can restart the work during this week for modifying the previous PR to use `ItemVerbatim`.

For closures, should i use [`syn::ExprVerbatim`](https://docs.rs/syn/0.14.5/syn/struct.ExprVerbatim.html)?

dtolnay:
I rebased #450 as f00a3efc7dceef0d57ae01c39a0a39293c80a776 and it will be releasing in 0.15.0.

PR #465, opened 2018-07-31
accept async functions and closures as verbatim

ubnt-intrepid:
This PR is a modification of #450 for avoiding breaking changes (https://github.com/dtolnay/syn/issues/395#issuecomment-409012200).

cc: @dtolnay, @steveklabnik

### DIFF
--- a/src/expr.rs
+++ b/src/expr.rs
@@ -1609,6 +1609,8 @@ pub mod parsing {
         |
         call!(expr_closure, allow_struct)
         |
+        call!(expr_unstable_async_closure, allow_struct)
+        |
         cond_reduce!(allow_block, syn!(ExprBlock)) => { Expr::Block }
         |
         // NOTE: This is the prefix-form of range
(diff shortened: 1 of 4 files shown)


steveklabnik:
Ah nice! I was just about to start looking into this; now I don’t have to :)

ubnt-intrepid:
Enabled the testing for async/await. Since async blocks are using in the test code, the support for its syntax is added at the same time.

I noticed that the async block cannot be parsed in the edition 2015. Therefore, in order to pass all the test cases, it seems necessary to detect the edition used in the test codes and set the parsing mode in `libsyntax`.

dtolnay:
Good catch. Would you be willing to look into detecting the `// edition:2018` comment in our test setup? Hopefully a basic `.contains("// edition:2018")` will do the trick.

dtolnay:
I ended up globally setting the edition in 5d871d99cbbeebb100b9b769e2bca41b5a748a2b.

dtolnay:
I released change in 0.14.6.

PR #466, opened 2018-08-03
Add an iter_mut method to Fields

mjbshaw:
Iterating over the fields is very useful, but I currently have a need to mutate the fields. I could implement this behavior locally in my project, but I think others might find this useful.


mjbshaw:
Thanks, done. Sorry I missed that one.

PR #498, opened 2018-09-22
Fix building on wasm32 target.

utkarshkukreti:
Hi,

#469 fixed building on wasm32 target but a later commit added code which broke it again. This PR changes the `cfg` attribute to match what #469 does.

Before:

```
$ cargo build --target wasm32-unknown-unknown
   Compiling syn v0.15.4 (file:///.../syn)
error[E0432]: unresolved import `proc_macro`
  --> src/export.rs:16:9
   |
16 | pub use proc_macro::TokenStream;
   |         ^^^^^^^^^^ Maybe a missing `extern crate proc_macro;`?

error: aborting due to previous error

For more information about this error, try `rustc --explain E0432`.
error: Could not compile `syn`.
```

After:

```
$ cargo build --target wasm32-unknown-unknown
   Compiling syn v0.15.4 (file:///.../syn)
    Finished dev [unoptimized + debuginfo] target(s) in 1.90s
```

I've tried compiling with all the feature combinations that are in `.travis.yml` with wasm32-unknown-unknown and they all work.

Not sure how I can add wasm32 target to travis.yml. I can look into it if you want.

### DIFF
--- a/src/export.rs
+++ b/src/export.rs
@@ -12,7 +12,10 @@ pub use proc_macro2::{Span, TokenStream as TokenStream2};
 
 pub use span::IntoSpans;
 
-#[cfg(feature = "proc-macro")]
+#[cfg(all(
+    not(all(target_arch = "wasm32", target_os = "unknown")),
+    feature = "proc-macro"
+))]
 pub use proc_macro::TokenStream;
 
 #[cfg(feature = "printing")]


dtolnay:
Thanks, I released this fix in 0.15.5 and added a wasm32-unknown-unknown builder in #500 that should catch failures like this.

utkarshkukreti:
@dtolnay thanks for merging so fast and the new travis config!

What would be the best way to ask questions about syn/quote (like how to do $x)? Should I ask in #rust on IRC or is there a better place?

utkarshkukreti:
Thank you, I will. I have one quick question: is there a better way to convert an `&str` to a literal string expression than this?:

```
syn::Expr::Lit(syn::ExprLit {
    attrs: Vec::new(),
    lit: syn::Lit::Str(syn::LitStr::new(str, proc_macro2::Span::call_site())),
})
```

PR #503, opened 2018-09-24
[WIP] custom_punctuation!

### LINKED ISSUE #502: Provide macro to define custom punctuation, analogous to custom keywords
For example a function-like macro may want to use `<=>` as punctuation. Parsing this as `Token![<]` `Token![=]` `Token![>]` would be incorrect because it does not enforce that the tokens are Joint.

CAD97:
Analog to `custom_keyword!`, `custom_punctuation!` allows a user of the library to treat a sequence symbols as one token.

Only loosely tested (to compile) so far, and using `CustomPunctuation` the same as `CustomKeyword` results in conflicting implementations of both `token::Sealed` and `token::Token`, as the two traits could both be implemented on the same type. For demonstrative purposes, `CustomKeyword` has been commented out.

The approach that will work for both traits will probably be making `token::Token` a supertrait and exposing `token::Sealed` in `$crate::export` to allow our macros to implement it.

Closes #502.


dtolnay:
CustomKeyword is private, you can rename it and use the same trait for both custom keywords and custom punctuation. That way only one blanket impl is necessary.

Here is a macro to compute length of a custom punctuation without str::len:

```rust
macro_rules! len {
    (<=) => { 2 };
    (>) => { 1 };
    /* ... */
    ($first:tt $($rest:tt)+) => {
        len!($first) $(+ len!($rest))*
    };
}

fn main() {
    let spans = [0; len!(<=>)];
    println!("{}", spans.len());
}
```

CAD97:
I _think_ the macro will still have to take punctuation in a string, though, as `stringify!(<=>)` is `"<= >"`. Or I could modify the `punct` helpers to skip over whitespace in the string if the ergonomy of the `custom_punctuation!` is more important.

The other idea I had was to have a `struct CustomPunctuation(&'static [char])` instead of using a trait.

(Given that this macro is only really useful in functionlike proc_macro town, we _could_ write it as a functionlike proc_macro and only expose it on Rust >= 1.29..... but that sounds like a bad idea and way too overkill.)

<details><summary>
I've not access to my development machine at this exact moment so I'm dropping this new sketch here:
</summary>

```rust
trait CustomToken<I: CustomIdent, P: CustomPunct> {
    fn token_str() -> &'static str;
    fn display() -> &'static str;
    fn spans(&self) -> &[Span]
}

// Keyword as before
impl CustomToken for Keyword {
    fn token_str() -> &'static str { stringify!($ident) }
    fn display() -> &'static str { concat!("`", stringify!($ident), "`") }
    fn spans(&self) -> &[Span] { &[self.span] }
}

macro custom_punctuation!( Punct($symbol:tts) ) {

struct Punct {
    pub spans: [Span; len!($symbol)];
}

impl CustomToken for Punct {
    fn token_str() -> &'static str { stringify!($symbol) } // WRONG
    fn display() -> &'static str { concat!("`", stringify!($symbol), "`") } // WRONG
    fn spans(&self) -> &[Span] { &self.spans }
}
impl Parse for Punct {
    fn parse(input) -> Result<Self> {
        // must expose parsing::punct somehow (or duplicate)
        Ok(Token { spans: parsing::punct(input, Self::token_str())? })
    }
}
impl ToTokens for Punct {
    fn to_tokens(&self, &mut tokens) {
        // must expose printing::punct somehow (or duplicate)
        printing::punct(Self::token_str(), self.spans(), tokens)
    }
}

}
```
</details><br/>

Side thought: might it make sense to robust-ify these macros a bit more and use them internally as well? Or at the very least some of the internal bits should be reusable between tokens that do and don't go through `CustomToken`.

```rust
macro custom_keyword!($meta $kw) {
    impl_custom_token!($meta $kw);
    impl_token_parse!($kw);
    // etc
}

macro internal_keyword!($meta $kw) {
    impl_internal_token!($meta $kw);
    impl_token_parse!($kw);
    // etc
}
```

(Just using macro keyword for compactness)

dtolnay:
Here's a stringify without spaces:

```rust
macro_rules! stringify_punct {
    ($($tt:tt)*) => {
        concat!($(stringify!($tt)),*)
    };
}

fn main() {
    let s: &'static str = stringify_punct!(<=>);
    println!("{:?}", s);
}
```

CAD97:
I'm not sure exactly where things should end up; I just chose arbitrarily to group stuff as it is in the above commit. `test_rustc_precedence` is failing on my end (%1 is not a valid Win32 application), so I'm letting Travis test.

CAD97:
Alright, this is working (modulo the use of `pub(restricted)` syntax which breaks the 1.15 build). I rebased down to one commit since it's one step and most of the commits were just "try fixes".

I just need to move stuff to the correct location, which will probably be just to not use `$crate::custom_token::punct` but rather `#[doc(hidden)] pub` the helpers in `$crate::token`, and use those directly. I'm also open to inlining the helpers manually as is (still) done for `custom_keyword`); which would you prefer @dtolnay?

src/custom_token.rs
```diff
@@ -0,0 +1,534 @@
...
+
+// Not public API.
+#[cfg(not(feature = "extra-traits"))]
+#[doc(hidden)]
+#[macro_export]
+macro_rules! impl_extra_traits_for_custom_punctuation {
+    ($ident: ident, $($tt:tt)*) => {};
+}
+
+custom_punctuation!(PathSep, </>);
```
dtolnay: Please remove.

src/custom_token.rs
```diff
@@ -0,0 +1,534 @@
...
+///
+/// - Field access to its spans — `let spans = path_separator_token.spans`
+///
+/// [Peeking]: parse/struct.ParseBuffer.html#method.peek
+/// [Parsing]: parse/struct.ParseBuffer.html#method.parse
+/// [Printing]: https://docs.rs/quote/0.6/quote/trait.ToTokens.html
+/// [`Span`]: struct.Span.html
+///
+/// # Example
+// TODO: actual example
```
dtolnay: This is going to need an actual example. Please come up with a reasonably fleshed out compilable example like we have for `custom_keyword!`.

src/custom_token.rs
```diff
@@ -0,0 +1,534 @@
...
+///
+/// - [Peeking] — `input.peek(punct::LeftRightArrow)`
+///
+/// - [Parsing] — `input.parse::<punct::LeftRightArrow>()?`
+///
+/// - [Printing] — `quote!( ... #path_separator_token ... )`
+///
+/// - Construction from (a) [`Span`]\(s) — `let path_separator_token = punct::LeftRightArrow(sp)`
+///
+/// - Field access to its spans — `let spans = path_separator_token.spans`
```
dtolnay: Let's be consistent and use `left_right_arrow` rather than `path_separator_token` in the above 3 lines.

src/token.rs
```diff
@@ -139,7 +139,7 @@ pub trait Token: private::Sealed {
 }
 
 #[cfg(feature = "parsing")]
-mod private {
+pub(crate) mod private {
```
dtolnay: Please solve this in a way that preserves support for rustc 1.15.

src/custom_token.rs
```diff
@@ -0,0 +1,534 @@
...
+    (;)      => ( 1usize );
+    (<<)     => ( 2usize );
+    (<<=)    => ( 3usize );
+    (>>)     => ( 2usize );
+    (>>=)    => ( 3usize );
+    (*)      => ( 1usize );
+    (-)      => ( 1usize );
+    (-=)     => ( 2usize );
+    (~)      => ( 1usize );
+    (_)      => ( 1usize );
```
dtolnay: Let's disallow underscore in custom punctuation. The compiler has transitioned to treating `_` as an identifier rather than punctuation because it is most commonly used in places where you would otherwise see an identifier, and it does not combine with other punctuation.

src/custom_token.rs
```diff
@@ -0,0 +1,534 @@
...
+    fn display() -> &'static str {
+        <Self as CustomToken>::display()
+    }
+}
+
+// Not public API.
+#[doc(hidden)]
+#[macro_export]
+macro_rules! punct_len {
+    ()       => ( 0usize );
```
dtolnay: Doesn't look like this rule is ever used. Please remove it is not used.

src/custom_token.rs
```diff
@@ -0,0 +1,534 @@
...
+    (<<=)    => ( 3usize );
+    (>>)     => ( 2usize );
+    (>>=)    => ( 3usize );
+    (*)      => ( 1usize );
+    (-)      => ( 1usize );
+    (-=)     => ( 2usize );
+    (~)      => ( 1usize );
+    (_)      => ( 1usize );
+    ($tt:tt) => ( compile_error!("Punctuation `{}` is not supported", stringify!($tt)) );
+    ($head:tt $($tail:tt)+) => ( punct_len!($head) + punct_len!($($tail)+) );
```
dtolnay: I would write this as follows. Nobody should be writing custom punctuation with 64+ characters but why mess with recursion when not necessary.

```rust
punct_len!($head) $(+ punct_len!($tail))*
```

src/custom_token.rs
```diff
@@ -0,0 +1,534 @@
...
+/// The generated syntax tree node supports the following operations just like
+/// any built-in punctuation token.
+///
+/// - [Peeking] — `input.peek(punct::LeftRightArrow)`
+///
+/// - [Parsing] — `input.parse::<punct::LeftRightArrow>()?`
+///
+/// - [Printing] — `quote!( ... #path_separator_token ... )`
+///
+/// - Construction from (a) [`Span`]\(s) — `let path_separator_token = punct::LeftRightArrow(sp)`
```
dtolnay: Show as a separate bullet point how to construct punctuation from multiple spans.

src/custom_token.rs
```diff
@@ -0,0 +1,534 @@
...
+        impl_clone_for_custom_punctuation!($ident, $($tt)*);
+        impl_extra_traits_for_custom_punctuation!($ident, $($tt)*);
+    };
+}
+
+// Not public API.
+#[cfg(feature = "parsing")]
+#[doc(hidden)]
+#[macro_export]
+macro_rules! impl_parsing_for_custom_punctuation {
```
dtolnay: Some of these new macros will require `#[macro_export(local_inner_macros)]` in order to support 2018 edition. Please try `custom_punctuation!` locally in a 2018 crate and confirm that it works.

src/custom_token.rs
CAD97: (whoops, still in there from testing)

src/custom_token.rs
CAD97: Leftovers from a previous iteration 😅

CAD97:
That should address everything!

src/custom_token.rs
```diff
@@ -0,0 +1,541 @@
...
+///
+/// // expr </> expr </> expr ...
+/// struct PathSegments {
+///     segments: Punctuated<Expr, punct::PathSeparator>,
+/// }
+///
+/// impl Parse for PathSegments {
+///     fn parse(input: ParseStream) -> Result<Self> {
+///         PathSegments {
+///             segments: Punctuated::parse_separated_nonempty(input)?,
```
dtolnay: I don't think this will do what you want unfortunately. Like in macro_rules, the Expr parser will expect that binary operators are followed by a rhs that is an expression. Indeed when I tried this code:

```console
error: expected expression
 --> src/main.rs:4:13
  |
4 | demo!(expr </> expr </> expr);
  |             ^
```

src/custom_token.rs
```diff
@@ -0,0 +1,541 @@
...
+/// }
+///
+/// // expr </> expr </> expr ...
+/// struct PathSegments {
+///     segments: Punctuated<Expr, punct::PathSeparator>,
+/// }
+///
+/// impl Parse for PathSegments {
+///     fn parse(input: ParseStream) -> Result<Self> {
+///         PathSegments {
```
dtolnay: This is going to need an `Ok(`...`)` around.

```console
error[E0308]: mismatched types
  --> src/custom_token.rs:390:9
   |
19 |       fn parse(input: ParseStream) -> Result<Self> {
   |                                       ------------ expected `std::result::Result<PathSegments, syn::parse::Error>` because of return type
20 | /         PathSegments {
21 | |             segments: Punctuated::parse_separated_nonempty(input)?,
22 | |         }
   | |_________^ expected enum `std::result::Result`, found struct `PathSegments`
   |
   = note: expected type `std::result::Result<PathSegments, syn::parse::Error>`
              found type `PathSegments`
```

dtolnay:
Still working on those two fixes?

CAD97:
Over in https://github.com/rust-lang/gll/pull/77 I discovered that the inbuilt proc_macro doesn't properly maintain `joint`-ness of `Punct`, so this is probably dead in the water until that's fixed upstream. (It still works with proc_macro2, just not proc_macro. Exposing this and having behavior differ depending on backend seems like the worst option, sadly.)

dtolnay:
It works fine unless you've hit https://github.com/rust-lang/rust/issues/43081. For example try writing `m!{%%}` in [dev/main.rs](https://github.com/dtolnay/syn/blob/0.15.31/dev/main.rs) and it parses as:

```rust
Macro {
    path: Path {
        leading_colon: None,
        segments: [
            PathSegment {
                ident: Ident {
                    ident: "m",
                    span: #0 bytes(94..95),
                },
                arguments: None,
            },
        ],
    },
    bang_token: Bang,
    delimiter: Brace(
        Brace,
    ),
    tts: TokenStream [
        Punct {
            ch: '%',
            spacing: Joint,
            span: #0 bytes(97..98),
        },
        Punct {
            ch: '%',
            spacing: Alone,
            span: #0 bytes(98..99),
        },
    ],
}
```

CAD97:
You're right, there's still a roundtrip through strings in the library >.<

PR #511, opened 2018-10-01
Check for invalid spans, and if so, use Span::call_site() instead

### LINKED ISSUE #508: Span information is not preserved for enums with variants that have named fields
Example:

```rust
enum E {
    Variant {
        field: ()
    }
}
```

Attempting to use the span like this:
```rust
quote_spanned!(input.span()=> compile_error!("Bad stuff");)
```
(where input can be `DeriveInput` or `ItemEnum`) the following will be printed as an error:
```
error: Bad stuff

error: aborting due to previous error

error: Could not compile `syn_enum_span`.

To learn more, run the command again with --verbose.
```

This obviously does not provide any useful information to the user: not even a filename is shown.

An example project is available here: 
[syn_enum_span.tar.gz](https://github.com/dtolnay/syn/files/2431035/syn_enum_span.tar.gz)

My version of rustc: `rustc 1.29.1 (b801ae664 2018-09-20)`

Laaas:
This checks the debug output of spans generated in Spanned::span
(quite a hack, yes..), and if it ends with "bytes(0..0)",
it assumes that it is invalid.

This fixes #508.

### DIFF
--- a/src/spanned.rs
+++ b/src/spanned.rs
@@ -110,6 +110,18 @@ mod private {
     impl<T: ToTokens> Sealed for T {}
 }
 
+// FIXME: This shouldn't be required, since optimally
+// spans should never be invalid. This can probably be removed when
+// https://github.com/rust-lang/rust/issues/43081 is resolved.
+fn check_invalid_span(span: Span) -> Option<Span> {
+    let debug = format!("{:?}", span);
+    if debug.ends_with("bytes(0..0)") {
+        None
+    } else {
+        Some(span)
+    }
+}
+
 impl<T> Spanned for T
 where
     T: ToTokens,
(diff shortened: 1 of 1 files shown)


src/spanned.rs
```diff
@@ -110,6 +110,18 @@ mod private {
     impl<T: ToTokens> Sealed for T {}
 }
 
+// FIXME: This shouldn't be required, since optimally
+// spans should never be invalid. This can probably be removed when
+// https://github.com/rust-lang/rust/issues/43081 is resolved.
+fn check_invalid_spans(span: Span) -> Span {
+    let debug = format!("{:?}", span);
+    if &debug[debug.len() - 11..] == "bytes(0..0)" {
```
dtolnay: I would prefer to use https://doc.rust-lang.org/std/primitive.str.html#method.ends_with to avoid a panic in the case that the debug format for Span changes to a shorter format.

src/spanned.rs
```diff
@@ -130,7 +142,8 @@ where
                 span = joined;
             }
         }
-        span
+
+        check_invalid_spans(span)
```
dtolnay: This check should happen above in the loop so that a syntax tree node can still have a valid span even if there are fragments interpolated into it that do not have valid spans. Let's skip any invalid spans when doing the joins.

src/spanned.rs
Laaas: That is a very good point.

Laaas:
@dtolnay the necessary changes have been made

src/spanned.rs
```diff
@@ -120,17 +132,18 @@ where
...
         };
         for tt in iter {
-            if let Some(joined) = span.join(tt.span()) {
+            if let Some(joined) = span.join(check_invalid_span(tt.span())) {
                 span = joined;
             }
         }
-        span
+
+        check_invalid_span(span) // Don't know if this is needed?
```
dtolnay: I don't know either. It depends whether joining two valid spans can produce an invalid span. Possibly, in the case that the input spans are not contiguous? I am fine with keeping this line.

src/spanned.rs
```diff
@@ -120,17 +132,18 @@ where
...
         let mut span = match iter.next() {
-            Some(tt) => tt.span(),
+            Some(tt) => check_invalid_span(tt.span()),
             None => {
                 return Span::call_site();
             }
         };
         for tt in iter {
-            if let Some(joined) = span.join(tt.span()) {
+            if let Some(joined) = span.join(check_invalid_span(tt.span())) {
```
dtolnay: Invalid spans need to be filtered out, not replaced with call_site. The behavior of Spanned::span should be to join all the valid spans in the input.

If there were no valid spans at all then we can return call_site.

Laaas:
@dtolnay I think it's better now

Laaas:
nvm I didn't realise that the special case isn't checked with cargo check

Laaas:
I don't think it's tested either

Laaas:
Well I haven't tested it @dtolnay since I get a stack overflow each time I try to test syn with all features enabled (with and without this patch), but both functions compile.

dtolnay:
What exact command are you running that has a stack overflow, and what operating system?

Laaas:
running arch linux, set my stack limit to 32 MiB. Couldn't get it higher due to the hard limit, and didn't bother changing it.

PR #519, opened 2018-10-21
Leverage language-level deprecation attributes

hcpl:
Also show deprecations in docs to inform users about this fact.


hcpl:
Okay, thanks for explanation!

PR #536, opened 2018-11-11
Make parse errors implement Send and Sync

### LINKED ISSUE #534: Mention Sync in documentation of syn::parse::Error
This comes up when converting Syn's errors to Failure's errors. Within a procedural macro you should never do this because it loses the ability to emit the error to the compiler in a way that points to the correct range of input tokens like a native compiler error.

Outside of a procedural macro, such as in a build script, you don't have that ability to begin with so you can write `failure::err_msg(err.to_string())`.

Suggested in #497.
Mentioning @gnzlbg.

### LINKED ISSUE #535: Investigate storing Span behind a Sync pointer so that Error can be Sync
Ideally the behavior that we want is an Error on the same thread that it was constructed on should behave like it has a span (the one passed when constructing the error), but the same error accessed from a different thread should behave like its span is Span::call_site(). I am not sure if this is possible.

The compiler's Span is an index into a threadlocal interner to keep the size small which is why reading the same Span from a different thread is not meaningful.

Mentioning @gnzlbg.

dtolnay:
Fixes #535. Closes #534.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -15,6 +15,7 @@ use proc_macro2::{
 };
 
 use buffer::Cursor;
+use thread::ThreadBound;
 
 /// The result of a Syn parser.
 pub type Result<T> = std::result::Result<T, Error>;
@@ -26,11 +27,19 @@ pub type Result<T> = std::result::Result<T, Error>;
 /// [module documentation]: index.html
 ///
 /// *This type is available if Syn is built with the `"parsing"` feature.*
-#[derive(Debug, Clone)]
+#[derive(Debug)]
 pub struct Error {
-    span: Span,
+    // Span is implemented as an index into a thread-local interner to keep the
+    // size small. It is not safe to access from a different thread. We want
+    // errors to be Send and Sync to play nicely with the Failure crate, so pin
+    // the span we're given to its original thread and assume it is
+    // Span::call_site if accessed from any other thread.
+    span: ThreadBound<Span>,
     message: String,
 }
+
+#[cfg(test)]
+struct _Test where Error: Send + Sync;
 
 impl Error {
     /// Usually the [`ParseStream::error`] method will be used instead, which
@@ -70,13 +79,16 @@ impl Error {
     /// ```
     pub fn new<T: Display>(span: Span, message: T) -> Self {
         Error {
-            span: span,
+            span: ThreadBound::new(span),
             message: message.to_string(),
         }
     }
 
     pub fn span(&self) -> Span {
-        self.span
+        match self.span.get() {
+            Some(span) => *span,
+            None => Span::call_site(),
+        }
     }
 
     /// Render the error as an invocation of [`compile_error!`].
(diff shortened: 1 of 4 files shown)


src/thread.rs
```diff
@@ -0,0 +1,66 @@
...
+            if thread::current().id() == self.thread_id {
+                Some(&self.value)
+            } else {
+                None
+            }
+        }
+    }
+}
+
+#[cfg(not(syn_can_use_thread_id))]
```
mystor: I think we can do better than this if we don't have `std::thread::Thread::id`, just with some performance overhead. This is safe, as we don't implement `Send` and `Sync` on the type, but perhaps unfortunate.

We can "emulate" ThreadId's behaviour using atomics and thread locals. If you decide not to bother with that, please add a comment explaining what restrictions are put on people using old rustc versions ;-).

```rust
#[cfg(syn_can_use_thread_id)]
mod thread_id {
    use std::thread;

    pub use std::thread::ThreadId;
    pub fn current() -> ThreadId {
        thread::current().id()
    }
}

#[cfg(not(syn_can_use_thread_id))]
mod thread_id {
    use std::sync::atomic::{AtomicUsize, Ordering, ATOMIC_USIZE_INIT};

    static NEXT_THREAD_ID: AtomicUsize = ATOMIC_USIZE_INIT;
    thread_local! {
        static THREAD_ID: usize = NEXT_THREAD_ID.fetch_add(1, Ordering::Relaxed);
    }

    pub type ThreadId = usize;
    pub fn current() -> ThreadId {
        THREAD_ID.with(|id| id)
    }
}
```

src/thread.rs
```diff
@@ -0,0 +1,66 @@
+use std::fmt::{self, Debug};
+
+#[cfg(syn_can_use_thread_id)]
+mod imp {
+    use std::thread::{self, ThreadId};
+
+    /// ThreadBound is a Sync-maker and Send-maker that allows accessing a value
+    /// of type T only from the original thread on which the ThreadBound was
+    /// constructed.
+    pub struct ThreadBound<T> {
```
mystor: I wonder if this should be split out into a separate crate rather than used internally - but *shrug*. It's not super generally useful & that might have some build overhead.

dtolnay: Might as well make a crate for it. I can take care of that this week. I will check what the impact on build time turns out to be but I am guessing it will be noticeable enough that we shouldn't use the new crate in Syn.

src/thread.rs
dtolnay: Nice! Done.

src/thread.rs
```diff
@@ -0,0 +1,71 @@
...
+}
+
+#[cfg(not(syn_can_use_thread_id))]
+mod thread_id {
+    use std::sync::atomic::{AtomicUsize, Ordering, ATOMIC_USIZE_INIT};
+
+    thread_local! {
+        static THREAD_ID: usize = {
+            static NEXT_THREAD_ID: AtomicUsize = ATOMIC_USIZE_INIT;
+            NEXT_THREAD_ID.fetch_add(1, Ordering::Relaxed)
```
dtolnay: @mystor why is Ordering::Relaxed correct here?

src/error.rs
```diff
@@ -26,12 +27,20 @@ pub type Result<T> = std::result::Result<T, Error>;
...
 /// *This type is available if Syn is built with the `"parsing"` feature.*
-#[derive(Debug, Clone)]
+#[derive(Debug)]
 pub struct Error {
-    span: Span,
+    // Span is implemented as an index into a thread-local interner to keep the
+    // size small. It is not safe to access from a different thread. We want
+    // errors to be Send and Sync to play nicely with the Failure crate, so pin
+    // the span we're given to its original thread and assume it is
+    // Span::call_site if accessed from any other thread.
```
gnzlbg: > assume it is `Span::call_site` if accessed from any other thread.

Wouldn't this produce worse error messages when passed to another thread ?

dtolnay: Yes it will. The compiler's source locations are all confined to one thread so there is no way around that.

gnzlbg: Can this PR also modify the docs of the `Error` type with this information ?

src/thread.rs
dtolnay: Thanks! Great explanation. I copied what you wrote into a comment.

PR #539, opened 2018-11-24
Export syn::Error and syn::Result from crate root

dtolnay:
Previously accessible as `syn::parse::Error` and `syn::parse::Result`.

I haven't fully made up my mind about this, but these have turned into a fairly nice general-purpose proc macro error type beyond just parsing. For example a derive macro that accepts attributes may want to use `syn::Error` as the error type for reporting a nonsensical application of some attribute, which is not exactly a parsing issue.

```rust
#[derive(Serialize)]
#[serde(untagged)]
struct S;
```

```console
error: #[serde(untagged)] can only be used on enums
 --> src/lib.rs:5:9
  |
5 | #[serde(untagged)]
  |         ^^^^^^^^
```

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -565,7 +565,6 @@ mod span;
 #[cfg(all(any(feature = "full", feature = "derive"), feature = "printing"))]
 mod print;
 
-#[cfg(feature = "parsing")]
 mod thread;
 
 ////////////////////////////////////////////////////////////////////////////////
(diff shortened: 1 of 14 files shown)


src/lib.rs
```diff
@@ -579,7 +579,7 @@ struct private;
 #[cfg(feature = "parsing")]
 mod error;
 #[cfg(feature = "parsing")]
```
mystor: Hmm, if we want these to be generally available, should they still stay behind the parsing feature? Perhaps a different feature might be relevant to use here.

dtolnay: Good call. These types don't affect compile time so it should be okay to provide them even with `default-features = false`.

mystor:
It'd also probably be a good idea to change the paths used in tests to match the new export paths :-)

PR #556, opened 2018-12-31
Improve parsing documentation

jethrogb:


jethrogb:
Why even have the `impl<F, T> Parser for F where F: FnOnce(ParseStream) -> Result<T>` if you don't want people to use it? From the example I added in #557 you can see that it's far more convenient to write small parsers that way (compared to manually implementing `Parse`).

jethrogb:
Also, why did you not include “The easiest way to create a `ParseStream` is to use the `Parser` trait”? It took me forever to figure that part out.

dtolnay:
Regarding how to create a ParseStream -- I disagreed with your wording and wrote a more complete explanation in 028a7d74f2fd859dfa654a213527b3d8b1947e14.

Regarding using the Parser impl -- you are correct that I don't want people to use it. Regarding why even have it -- it is a low-level escape hatch that is necessary on rare occasions. I try never to base design primarily on "far more convenient to write." I believe data structures with a Parse impl are easier to think about, read, and document.

PR #565, opened 2019-01-16
Make turbofish optional

dtolnay:
Implements https://github.com/rust-lang/rfcs/pull/2544 which is in proposed-FCP.

This simple implementation works but diagnostics are somewhat worse than before. Consider the following input in which the user is missing part of the second type parameter. The parser backtracks to reparse the first type parameter as an expression, concluding that it is supposed to be something like `&('a: loop {})` and suggesting that the user write the missing colon after their loop label.

We may be able to build out a system for collecting both errors and emitting whichever one happened after parsing the larger number of tokens.

#### Before:

```console
error: expected identifier
 --> src/main.rs:4:38
  |
4 |     let _ = f::<&'a fn(), for<'a> serde::>();
  |                                          ^
```

#### After, without turbofish:

```console
error: expected `:`
 --> src/main.rs:4:19
  |
4 |     let _ = f<&'a fn(), for<'a> serde::>();
  |                   ^^
```

### DIFF
--- a/src/expr.rs
+++ b/src/expr.rs
@@ -2261,7 +2261,7 @@ pub mod parsing {
             #[cfg(feature = "full")]
             let attrs = input.call(Attribute::parse_outer)?;
 
-            let (qself, path) = path::parsing::qpath(input, true)?;
+            let (qself, path) = path::parsing::qpath(input)?;
 
             Ok(ExprPath {
                 attrs: attrs,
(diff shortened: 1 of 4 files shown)


dtolnay:
Closing for now as https://github.com/rust-lang/rfcs/pull/2544 has stalled.

Issue #479, opened 2018-09-04
Public bridge between low and high level parser APIs.


mystor:
I had some code in a project where I was constructing a `TokenBuffer` from a `TokenStream` myself, and then using the `Cursor` to walk along it. It appears as though there is no public mechanism currently to get from a `TokenBuffer` or `Cursor` object to a `ParseBuffer`. It could be nice if some public API for this was provided beyond using `Parser::parse2` on a lambda.

dtolnay:
Thanks! You are right, this is not well supported now.

This is going to be tricky because `Parser::parse2` is actually what detects tokens that were never parsed -- even inside of nested groups. That's how we get away without any sort of checking for remaining tokens in impls like this one:

https://github.com/dtolnay/syn/blob/719ad5a1c1944c3b413179371f4a621c010eeb8f/src/ty.rs#L455-L465

Without some sort of explicit check that `content.is_empty()` at the end of this code, those "unexpected token" errors need to be triggered from Drop impls and/or from some central place afterward. And as Drop impls cannot return error, that central place ends up being `Parser::parse2`.

That means going from `Cursor` temporarily to `ParseStream` and then back to `Cursor` is not simple (or not obvious to me how to make it simple).

I would like to understand whether more of your code could be written in terms of `ParseStream`. If all of it can, that seems better. If some of it needs to remain using `Cursor`, can just those parts be performed using a call to [`ParseStream::step`](https://docs.rs/syn-next/0.15.0-rc3/syn/parse/struct.ParseBuffer.html#method.step)? The closure argument to `step` effectively has the same signature as `Synom::parse`. Either way this would let `ParseStream` own the entire process and perform the right checks for unparsed tokens.

mystor:
I was able to fairly easily rewrite the logic to work with `ParseStream` (It wasn't super performance sensitive, so I just used `ParseBuffer::fork`).

I didn't think of the bracketed-parser issues, so it makes sense why the wrapper method call is needed. I suppose I would like a `parse_cursor` mechanism which takes in a `Cursor` and has the same semantics as `parse2(F, cursor.token_stream())`, but without the token copies. 

e.g.
```rust
trait Parser : Sealed {
    fn parse_cursor(self, cursor: Cursor) -> Result<T>;

    fn parse2(self, tokens: TokenStream) -> Result<T> {
        let buf = TokenBuffer::new2(tokens);
        self.parse_cursor(buf.begin())
    }
}

impl Parser for ... {
[...]

PR #567, opened 2019-01-20
Add unsafe ParseBuffer::join, the opposite of fork

CAD97:
Addresses (some of?) #479 by allowing speculative parsing without leaving the `ParseStream` land.

The method as is currently in the PR is unsafe, as it requires the user to guarantee that the fork is indeed forked from the parse stream it is being joined to. This could potentially be guaranteed by examining the state of the two parse streams' `Span`s, but doing so is not stable.

This could potentially be made to panic safely by making `self.scope` a `Rc` and using `Rc::ptr_eq` to make sure that the base and the fork are both using the same root `Rc`'d scope. The second commit (TBD) implements this approach.

### DIFF
--- a/src/parse.rs
+++ b/src/parse.rs
@@ -848,6 +848,121 @@ impl<'a> ParseBuffer<'a> {
         }
     }
 
+    /// Join this parse stream up with a forked parse stream.
+    ///
+    /// This is the opposite operation to [`ParseStream::fork`].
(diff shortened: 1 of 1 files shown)


CAD97:
(whoops, I forgot a word, fixing)
(reminder to self: always test locally first it's never just going to be right)

PR #574, opened 2019-01-24
add a u128 feature that makes LitInt::value return u128

### LINKED ISSUE #571: Make LitInt::value return u128 instead of u64
`u128` has been stable for quite some time. Yet when I want to get the value out of a `LitInt' I'm stuck with either accepting that literals larger than 64 bit are going to be truncated or I need to `quote!(_)` them and implement my own int literal parser. Both options are suboptimal.

So I propose changing the return type of `value()` to `u128`. This would constitute a breaking change, however, so I'd be ok with adding a `value_u128()` function instead (and should we need to keep compatibility with older Rust versions, we may hide it behind a feature flag).

llogiq:
This fixes #571


dtolnay:
Cargo cfgs are required to be additive which means they must not break code when enabled. Under this PR, a program like the following would break when the u128 feature is enabled by unrelated code.

```rust
fn main() {
    let _: fn(&syn::LitInt) -> u64 = syn::LitInt::value;
}
```

llogiq:
Ok, I changed the code to only include new methods with a `_u128` suffix.

dtolnay:
Thanks for the PR.

I would prefer to postpone any change to how literals are parsed until the next major version of Syn. Especially I don't think it is good that this PR would change large integers from being parsed as LitVerbatim to being parsed as LitInt on which the LitInt::value method simply truncates away the high bits.

Separately, I believe u128 should not play the role of "widest integer type" in public APIs so I would be opposed to changing LitInt::value from u64 to u128 in the future.

The Lit family of types is due for some fresh design work anyway. Likely this will end up being something like a `fn base10_digits(&self) -> String` method as the accessor for the integer value, which can then be parsed by str::parse. I would be willing to consider a PR that implements a base10_digits which works for an arbitrarily wide integer literal, without yet changing which literals get parsed as LitInt.

llogiq:
>  I don't think it is good that this PR would change large integers from being parsed as LitVerbatim to being parsed as LitInt

But only if the `u128` feature is activated.

> I believe u128 should not play the role of "widest integer type" in public APIs

Could you please help me understand your reasoning? As things stand, at least in `std`, u128 *is* the widest integer type available. Why shouldn't it be?

The way I see it, storing the value in some kind of canonical form is necessary to allow simple comparisons, hashes, etc. There are binary, octal, decimal and hexadecimal integer literals and all of them can contain an arbitrary number of underscores at arbitrary places. 

> fn base10_digits(&self)

That would give us such a canonical form. Why base 10? I note that this is wasteful compared to a base-256 `Vec<u8>`.

PR #578, opened 2019-02-08
Sync with Rust post C-variadic PR

dlrobertson:
rust-lang/rust#57760 "Support defining C compatible variadic
functions" changes the `FnDecl` and `FnSig` member `variadic` to
`c_variadic` and adds the `TyKind` variant `CVarArgs`. Update
`syn` to handle these changes.

Depends On: rust-lang/rust#57760

### DIFF
--- a/src/item.rs
+++ b/src/item.rs
@@ -1092,6 +1092,14 @@ pub mod parsing {
             let content;
             let paren_token = parenthesized!(content in input);
             let inputs = content.parse_terminated(FnArg::parse)?;
+            let variadic: Option<Token![...]> = match inputs.last() {
+                Some(punctuated::Pair::End(&FnArg::Captured(ArgCaptured {
+                    ty: Type::Verbatim(TypeVerbatim { ref tts }), ..
+                }))) => {
+                    parse2(tts.clone()).ok()
+                }
+                _ => None
+            };
 
             let output: ReturnType = input.parse()?;
             let where_clause: Option<WhereClause> = input.parse()?;
@@ -1114,7 +1122,7 @@ pub mod parsing {
                     paren_token: paren_token,
                     inputs: inputs,
                     output: output,
-                    variadic: None,
+                    variadic: variadic,
                     generics: Generics {
                         where_clause: where_clause,
                         ..generics
@@ -1179,7 +1187,27 @@ pub mod parsing {
         Ok(ArgCaptured {
             pat: input.parse()?,
             colon_token: input.parse()?,
-            ty: input.parse()?,
+            ty: match input.parse::<Token![...]>() {
+                Ok(dot3) => {
+                    let mut args = vec![
+                        TokenTree::Punct(Punct::new('.', Spacing::Joint)),
+                        TokenTree::Punct(Punct::new('.', Spacing::Joint)),
+                        TokenTree::Punct(Punct::new('.', Spacing::Alone)),
+                    ];
+                    let tokens = TokenStream::from_iter(
+                        args.into_iter().zip(&dot3.spans).map(|(mut arg, span)| {
+                            arg.set_span(*span);
+                            arg
+                        }
+                    ));
+                    Type::Verbatim(TypeVerbatim {
+                        tts: tokens
+                    })
+                }
+                Err(_) => {
+                   input.parse()?
+                }
+            },
         })
     }
 
@@ -2593,8 +2621,7 @@ mod printing {
 
     impl ToTokens for ForeignItemVerbatim {
         fn to_tokens(&self, tokens: &mut TokenStream) {
-            self.tts.to_tokens(tokens);
-        }
+            self.tts.to_tokens(tokens); }
     }
 
     impl ToTokens for MethodSig {
--- a/tests/clone.sh
+++ b/tests/clone.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-REV=e2f221c75932de7a29845c8d6f1f73536ad00c41
+REV=1999a2288123173b2e487865c9a04386173025f7
 
 set -euo pipefail
 cd "$(dirname "${BASH_SOURCE[0]}")"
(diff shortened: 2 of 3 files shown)


dlrobertson:
> Please do not rename the fields on FnDecl and TypeBareFn.

@dtolnay do you mean remove https://github.com/rust-lang/rust/pull/57760/commits/91cfa584b5e7c4f85d1eaac16d06f821439c2dc8 from [rust-lang/rust#57760](https://github.com/rust-lang/rust/pull/57760)? Or do you mean don't change https://github.com/dlrobertson/syn/blob/update_post_c_variadics_pr/src/item.rs#L707 and https://github.com/dlrobertson/syn/blob/update_post_c_variadics_pr/src/ty.rs#L74?

> Separately, please ensure that all the example code from the RFC parses successfully, including signatures like `fn func(fixed: u32, mut args: ...)`. The `...` type of the last parameter will need to be parsed as syn::Type::Verbatim.

I did  try to follow the RFC as close as I could. There are [run-make-fulldeps tests](https://github.com/dlrobertson/rust/blob/varargs1/src/test/run-make-fulldeps/c-link-to-rust-va-list-fn), [codegen tests](https://github.com/dlrobertson/rust/blob/varargs1/src/test/codegen/c-variadic.rs), and [ui tests](https://github.com/dlrobertson/rust/tree/varargs1/src/test/ui/c-variadic) all of which conform the RFC AFAIK. If you see some edge cases that I forgot to test, please ping me and I'll add some more.

Are there tests that should be added to this PR?

dtolnay:
I mean please do not change the field names in Syn's syntax tree. The compiler's syntax tree is not up to me.

I don't see any parser changes in this PR. I would expect new parsing code that handles parsing the new syntax. As is, the new syntax fails to parse.

```console
error: expected one of: `for`, parentheses, `fn`, `unsafe`, `extern`, identifier, `::`, `<`, square brackets, `*`, `&`, `!`, `impl`, `_`, lifetime
 --> src/main.rs:2:63
  |
2 |     pub unsafe extern "C" fn func(arg: T, arg2: T2, mut args: ...) {
  |                                                               ^^^
```

Please implement the relevant parsing logic and then confirm that Syn successfully parses rustc's test suite by testing with the commit hash in tests/clone.sh bumped to the commit hash of your rust-lang/rust PR.

dtolnay:
Here is the snippet that currently fails to parse: https://github.com/dtolnay/syn/commit/1f30c83e5867ea96ec1dbd9813bb51a2f4998277

dlrobertson:
> Here is the snippet that currently fails to parse: [1f30c83](https://github.com/dtolnay/syn/commit/1f30c83e5867ea96ec1dbd9813bb51a2f4998277)

:+1: thanks

dtolnay:
- Please preserve compatibility with rustc 1.15.
- I would prefer not to accept `...` anywhere that a type is parsed, only in a function argument. I would make the change on [this line](https://github.com/dtolnay/syn/blob/0.15.26/src/item.rs#L1182) to check for `Token![...]` before parsing a type.

src/item.rs
```diff
@@ -1179,7 +1190,13 @@ pub mod parsing {
         Ok(ArgCaptured {
             pat: input.parse()?,
             colon_token: input.parse()?,
-            ty: input.parse()?,
+            ty: if input.peek(Token![...]) {
+                Type::Verbatim(TypeVerbatim {
+                  tts: input.parse()?
```
dtolnay: I don't think this is correct -- tts is a TokenStream so you would be slurping up the entire rest of the input as that TokenStream.

Separately, please check that your indentation is 4 spaces per Rust style.

```rust
fn f(a: ..., b: u8) {}
```

```rust
File {
    shebang: None,
    attrs: [],
    items: [
        Fn(
            ItemFn {
                attrs: [],
                vis: Inherited,
                constness: None,
                unsafety: None,
                asyncness: None,
                abi: None,
                ident: Ident {
                    ident: "f",
                    span: #0 bytes(97..98)
                },
                decl: FnDecl {
                    fn_token: Fn,
                    generics: Generics {
                        lt_token: None,
                        params: [],
                        gt_token: None,
                        where_clause: None
                    },
                    paren_token: Paren,
                    inputs: [
                        Captured(
                            ArgCaptured {
                                pat: Ident(
                                    PatIdent {
                                        by_ref: None,
                                        mutability: None,
                                        ident: Ident {
                                            ident: "a",
                                            span: #0 bytes(99..100)
                                        },
                                        subpat: None
                                    }
                                ),
                                colon_token: Colon,
                                ty: Verbatim(
                                    TypeVerbatim {
                                        tts: TokenStream [
                                            Punct {
                                                ch: '.',
                                                spacing: Joint,
                                                span: #0 bytes(102..105)
                                            },
                                            Punct {
                                                ch: '.',
                                                spacing: Joint,
                                                span: #0 bytes(102..105)
                                            },
                                            Punct {
                                                ch: '.',
                                                spacing: Joint,
                                                span: #0 bytes(102..105)
                                            },
                                            Punct {
                                                ch: ',',
                                                spacing: Alone,
                                                span: #0 bytes(105..106)
                                            },
                                            Ident {
                                                ident: "b",
                                                span: #0 bytes(107..108)
                                            },
                                            Punct {
                                                ch: ':',
                                                spacing: Alone,
                                                span: #0 bytes(108..109)
                                            },
                                            Ident {
                                                ident: "u8",
                                                span: #0 bytes(110..112)
                                            }
                                        ]
                                    }
                                )
                            }
                        )
                    ],
                    variadic: None,
                    output: Default
                },
                block: Block {
                    brace_token: Brace,
                    stmts: []
                }
            }
        )
    ]
}
```

src/item.rs
```diff
@@ -1092,6 +1092,17 @@ pub mod parsing {
             let content;
             let paren_token = parenthesized!(content in input);
             let inputs = content.parse_terminated(FnArg::parse)?;
+            let variadic: Option<Token![...]> = match inputs.last() {
+                Some(punctuated::Pair::End(&FnArg::Captured(ArgCaptured {
+                    ty: Type::Verbatim(TypeVerbatim { ref tts }), ..
+                }))) => {
+                    match parse2::<token::Dot3>(tts.clone()) {
+                        Ok(tok) => Some(tok),
```
dtolnay: This could be written more compactly as `parse2(tts.clone()).ok()`.

dlrobertson: :+1: good point

src/item.rs
dlrobertson: Good catch! I'll update this and if I can figure out the tests I might add a test.

dtolnay:
Thanks! This looks good to me.

PR #580, opened 2019-02-15
Generate JSON file of all syn types

carllerche:
The JSON file can be used by third party libs to build tools that
consume `syn` types.

The JSON is generated to `syn.json`. [Link](https://github.com/carllerche/syn/blob/introspect-json/syn.json).

Follow up to #579.

### DIFF
--- a/codegen/Cargo.toml
+++ b/codegen/Cargo.toml
@@ -12,6 +12,10 @@ failure = "0.1"
 inflections = "1.1"
 proc-macro2 = "0.4"
 rustfmt-nightly = { git = "https://github.com/rust-lang-nursery/rustfmt" }
+serde = "1.0.87"
+serde_derive = "1.0.87"
+serde_json = "1.0.38"
+toml = "0.4.10"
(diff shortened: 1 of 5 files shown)


dtolnay:
I merged this with a few changes right after that make the json easier to follow (in my opinion). Let me know if any of those cause trouble for your use case.

carllerche:
LGTM. One issue I see is `Span` and `Ident` are now under `syn` it looks like? That probably should be fixed to be clasified as `proc_macro2`?

carllerche:
It comes from this: https://github.com/dtolnay/syn/blob/master/codegen/src/parse.rs#L15

dtolnay:
I think that happens because we currently generate `fold_ident` and `fold_span` methods but not `fold_literal` or `fold_token_stream`. I would be on board with reclassifying them as "proc_macro2" as long as the generated Rust code doesn't change.

carllerche:
Ok 👍 let me try

PR #596, opened 2019-03-10
Use 2018 quote::quote

nhynes:


dtolnay:
Uh-oh, looks like we are not shipping build.rs, which is why the version detection is not kicking in.

https://github.com/dtolnay/syn/blob/d9d5b25b9dbf79a893b18051c94392cc14e47863/Cargo.toml#L11

dtolnay:
I released 0.15.29 with the fix in #598. Thanks anyway!

PR #605, opened 2019-03-15
Parse paths in Meta

### LINKED ISSUE #597: Parse paths in Meta
[MWE](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=2977a11e2a94ea1e92ac2c76f99bb681)

The following is valid syntax, but is not parseable by `syn`. I'm not sure if this is expected behavior.
```rust
#[macro_use]
extern crate quote;

fn main() {
    let attr: syn::Attribute = syn::parse_quote!(#[derive(serde::Serialize)]);
    attr.parse_meta().unwrap();
}
```

azriel91:
Closes #597.

Now we can have:

```rust
#[foo(bar::Baz)] // bar::Baz is a `Meta::Path(Path)`
```

~~Bumped the version to `0.16.0` as this is a breaking change.~~
Got the tests going with `cargo insta`.
Couldn't find a `CHANGELOG.md`, not sure if you would like one.

---

Just realized, bumping `syn` to `0.16` in the whole ecosystem is an upcoming chore.

### DIFF
--- a/src/attr.rs
+++ b/src/attr.rs
@@ -184,18 +184,17 @@ impl Attribute {
     /// possible.
     #[cfg(all(feature = "parsing", not(feature = "clone-impls")))]
     pub fn parse_meta(&self) -> Result<Meta> {
-        let path = Self::clone_ident_path(&self.path)
-            .ok_or_else(|| {
-                #[cfg(all(feature = "parsing", feature = "printing"))]
-                let span = {
-                    use spanned::Spanned;
-                    self.path.span()
-                };
-                #[cfg(all(feature = "parsing", not(feature = "printing")))]
-                let span = proc_macro2::Span::call_site();
+        let path = Self::clone_ident_path(&self.path).ok_or_else(|| {
+            #[cfg(all(feature = "parsing", feature = "printing"))]
+            let span = {
+                use spanned::Spanned;
+                self.path.span()
+            };
+            #[cfg(all(feature = "parsing", not(feature = "printing")))]
+            let span = proc_macro2::Span::call_site();
 
-                Error::new(span, "path contained more than one segment")
-            })?;
+            Error::new(span, "path contained more than one segment")
+        })?;
 
         let parser = |input: ParseStream| parsing::parse_meta_after_path(path, input);
         parse::Parser::parse2(parser, self.tts.clone())
@@ -291,7 +290,9 @@ impl Attribute {
     ))]
     fn clone_ident_path(path: &Path) -> Option<Path> {
         if path.segments.len() == 1 {
-            path.segments.first().map(|pair| pair.value().ident.clone().into())
+            path.segments
+                .first()
+                .map(|pair| pair.value().ident.clone().into())
         } else {
             return None;
         }
@@ -316,7 +317,9 @@ fn nested_meta_item_from_tokens(tts: &[TokenTree]) -> Option<(NestedMeta, &[Toke
         // It does not cater for path Meta items with more than one segment.
         TokenTree::Ident(ref ident) => {
             if tts.len() >= 3 {
-                if let Some(meta) = Attribute::extract_name_value(ident.clone().into(), &tts[1], &tts[2]) {
+                if let Some(meta) =
+                    Attribute::extract_name_value(ident.clone().into(), &tts[1], &tts[2])
+                {
                     return Some((NestedMeta::Meta(meta), &tts[3..]));
                 }
             }
@@ -572,7 +575,7 @@ where
 pub mod parsing {
     use super::*;
 
-    use ext::PathExt;
+    use ext::IdentExt;
     use parse::{Parse, ParseStream, Result};
     #[cfg(feature = "full")]
     use private;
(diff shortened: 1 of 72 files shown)


Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "syn"
-version = "0.15.29" # don't forget to update html_root_url and syn.json
+version = "0.16.0" # don't forget to update html_root_url and syn.json
```
dtolnay: Please remove this from this PR. I will handle it separately.

src/attr.rs
```diff
@@ -578,9 +597,30 @@ pub mod parsing {
...
+        /// For backward compatibility, we will treat single segment paths as
+        /// `Meta::Word`s.
         fn parse(input: ParseStream) -> Result<Self> {
-            let ident = input.call(Ident::parse_any)?;
-            parse_meta_after_ident(ident, input)
+            // We assume `Path` values to contain `::` as single words are parsed as `Meta::Word`
+            // for backward compatibility.
+            if input.peek(Token![::]) || input.peek2(Token![::]) {
+                let path = input.call(Path::parse_mod_style)?;
+                Ok(Meta::Path(path))
```
dtolnay: What about `#[path::to = ""]` and `#[path::to("")]`?

dtolnay:
Thanks!

dtolnay:
Please remove the issue number from the individual commit messages because it spams #597 when commits are amended. Linking from the PR message is sufficient.

src/attr.rs
azriel91: hmm, earlier intentionally decided to leave those out. If we include that, there's two umbrella designs that are apparent:

* Expand the variants of `Meta` to have both the `Ident` and `Path` versions:

    ```rust
    ListIdent(MetaListIdent {
        pub ident: Ident,
        pub meta_list: MetaList,
    }),
    ListPath(MetaListPath {
        pub path: Path,
        pub meta_list: MetaList,
    })

    // and
    MetaList {
        pub paren_token: token::Paren,
        pub nested: Punctuated<NestedMeta, Token![,]>,
    }
    ```

* Or, replace `Ident` with `Path`

    ```rust
    List(MetaList {
        pub path: Path,
        pub paren_token: token::Paren,
        pub nested: Punctuated<NestedMeta, Token![,]>,
    })
    ```

Or something in between. The second option isn't exactly straightforward -- right now `Ident::parse_any` happily parses keywords, so substituting `Ident` with `Path` would mean figuring out:

* Should `Path` have a parse that accepts keywords?
* If yes, it should *likely* be a separate parse function -- we don't want to parse keywords everywhere.
* If no, then we are back to choosing one of those overall designs (or hybrid).

Is there a preference?

azriel91:
Okay, have removed the issue links in the commits (but the historical spam remains, sorry! Shall remember for next time).
The path in `List` / `NameValue` issue has high impact, so would like more experienced eyes on it ☝️.

src/attr.rs
dtolnay: I would like to keep Meta as an enum with three variants but use Path instead of Ident in each. The parsing for Meta should accept keywords in the path but the parsing can be implemented in attr.rs without exposing a new public parser function on Path.

azriel91:
Okay think it's done. Have updated and added tests, but would like to call out these specially:

* https://github.com/dtolnay/syn/pull/605/files#diff-b64ddf45091b7aea3c129600532d047c

    `interpret_meta` now accepts `path`s (changes previous behaviour)

* https://github.com/dtolnay/syn/pull/605/files#diff-4b9f3caeb7d15ba0422250fb31e55930R542

    using `parse_meta` means keywords are now accepted in attribute locations, e.g. `#[extern]`. Not sure if that's wanted.

azriel91:
CI failures:

* many failures are `Path` doesn't impl `Clone` (please post if you know why, I haven't investigated yet)
* Rust 1.15 doesn't support `pub(crate)`

Also, just realized I left a `// FIXME` in there.

dtolnay:
Thanks for sticking with this. I am on vacation this week so it may take longer to review.

- I mentioned earlier that I am not interested in exposing a new public parser on Path. The parsing here is specific to Meta and should be internal to Syn. Please find a way to write the parsing code in attr.rs without anything new public on Path or a new extension trait.

- I have not read the new implementation but for now I don't understand why stuff needs to be conditional on clone-impls. Conceptually, parsing a Meta does not seem to necessarily require cloning complicated types. It would be better if the behavior were independent of clone-impls.

azriel91:
Ah, sorry my mistake -- accidentally made the path-for-meta parsing public when saw that Rust 1.15 doesn't support `pub(crate)`. Have moved it into `attr.rs`.

Re:parsing being conditional on clone-impls, the previous comment should clarify, but don't worry about checking it until after vacation -- work can wait, life must not.

dtolnay:
Can you find any way to support multiple segment paths without a Clone impl on Path?

<br>

```rust
fn clone_path_of_attribute(path: &Path) -> Path {
    /* this seems feasible without clone-impls */
}

azriel91:
Consider it done ⚔️!

dtolnay:
Thanks! I will pull this in when making the next round of syntax tree breaking changes. Leaving #597 open to remember to do so.

If you need to be able to use this in the near term, I would recommend extracting the new Meta type into a crate that you can publish by itself.

PR #614, opened 2019-04-02
Parse implementation for ExprMatch

### LINKED ISSUE #601: `ExprMatch` does not have a `Parse` impl
In order to parse a `match` expression, you have to parse an expression and then check if it's a match expression. If you are expecting a match expression then just being able to parse that would give better errors and be simpler.

peterjoel:
Fixes #601


dtolnay:
Published in 0.15.30.

PR #568, opened 2019-01-20
Add ParseBuffer::advance_to, the opposite of fork

CAD97:
Addresses (some of?) #479 by allowing speculative parsing without leaving the ParseStream land.

The method in the first commit is unsafe, as it requires the user to guarantee that the fork is indeed forked from the parse stream it is being joined to. This could potentially be guaranteed by examining the state of the two parse streams' Spans, but doing so is not stable.

This is made to panic safely by making self.scope a Rc and using Rc::ptr_eq to make sure that the base and the fork are both using the same root Rc'd scope. The second commit implements this approach.

PR #619, opened 2019-04-13
impl Parse for UseTree

### LINKED ISSUE #618: impl Parse for UseTree
I believe the current private `syn::item::use_tree` is the only context to parse a `UseTree`, and it would be able to parse the `UseTree` for procedural macros that want to have a use tree in a different position than a full `ItemUse`.

https://github.com/dtolnay/syn/blob/f2f9a538858153a2a1bfacbf6098d71c5cb06965/src/item.rs#L993-L1038

At the very least a comment explaining why this is a "private impl" would be useful.

CAD97:
Closes #618 by exposing `impl Parse for UseTree`.

Diff is best viewed [without whitespace](https://github.com/dtolnay/syn/pull/619/files?w=1).

### DIFF
--- a/src/item.rs
+++ b/src/item.rs
@@ -984,56 +984,58 @@ pub mod parsing {
                 vis: input.parse()?,
                 use_token: input.parse()?,
                 leading_colon: input.parse()?,
-                tree: input.call(use_tree)?,
+                tree: input.parse()?,
                 semi_token: input.parse()?,
             })
         }
     }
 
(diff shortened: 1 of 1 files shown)


CAD97:
Fixed. (Thought I could get away without local testing. I was wrong.)

dtolnay:
Published in 0.15.31.

Still eager to have #568 when you get back to it.

PR #625, opened 2019-04-22
Extension method to unraw an ident

### LINKED ISSUE #478: Get the bare name of a raw Ident
Currently, all to string conversions on `Ident` return the ident verbatim, e.g. `r#move` as `r#move`. It would be very useful to have a way to get the bare name, e.g. `move`.

This would be useful e.g. for creating wrappers in pyo3:

```rust
fn function_wrapper_ident(name: &syn::Ident) -> syn::Ident {
    syn::Ident::new(&format!("__pyo3_get_function_{}", name), Span::call_site())
}
```

This version errors when given `r#move` with `"__pyo3_get_function_r#move" is not a valid identifier`. It's possible to work around this by replacing `r#` with an empty string, but I'd prefer it if syn provided a method for that.

### LINKED ISSUE #603: [feature request] r#-aware apis
Currently syn seems be largely ignorant of the existence of `r#`. This is correct in some situations, and incorrect in others. It would be good if the API of Ident (and others?) be adjusted/augmented to make it easier to correctly handle this.

Some notable examples with an `Ident("r#match")`:

* prefixing: `new_name = format!("prefix_{}", ident);` => `"prefix_r#match"` **(wrong)**
* exporting: `fn_name = ident.to_string();` => `void r#match(int32_t x, bool y);` **(wrong)**
* generating rust: `format!("struct {}", ident);` => `struct r#match` **(right!)**

At minimum having a to_unescaped_string() fn that strips r#'s would be nice, but I'm not an expert in the domain so it might be possible there's a more elegant/helpful solution here.

dtolnay:
```rust
trait IdentExt {
    fn unraw(&self) -> Ident;
}
```

Strips the raw marker `r#`, if any, from the beginning of an ident.

- unraw(`x`) = `x`
- unraw(`move`) = `move`
- unraw(`r#move`) = `move`

### Example

In the case of interop with other languages like Python that have a different set of keywords than Rust, we might come across macro input that involves raw identifiers to refer to ordinary variables in the other language with a name that happens to be a Rust keyword.

The function below appends an identifier from the caller's input onto a fixed prefix. Without using `unraw()`, this would tend to produce invalid identifiers like `__pyo3_get_r#move`.

```rust
use proc_macro2::Span;
use syn::Ident;
use syn::ext::IdentExt;

fn ident_for_getter(variable: &Ident) -> Ident {
    let getter = format!("__pyo3_get_{}", name.unraw());
    Ident::new(&getter, Span::call_site())
}
```

---

Fixes #478. Fixes #603.

### DIFF
--- a/src/ext.rs
+++ b/src/ext.rs
@@ -6,9 +6,10 @@ use proc_macro2::Ident;
 
 use parse::{ParseStream, Result};
 
-/// Additional parsing methods for `Ident`.
+/// Additional methods for `Ident` not provided by proc-macro2 or libproc_macro.
 ///
-/// This trait is sealed and cannot be implemented for types outside of Syn.
+/// This trait is sealed and cannot be implemented for types outside of Syn. It
+/// is implemented only for `proc_macro2::Ident`.
 ///
 /// *This trait is available if Syn is built with the `"parsing"` feature.*
 pub trait IdentExt: Sized + private::Sealed {
@@ -16,6 +17,8 @@ pub trait IdentExt: Sized + private::Sealed {
     ///
     /// This is useful when parsing a DSL which allows Rust keywords as
     /// identifiers.
+    ///
+    /// # Example
     ///
     /// ```edition2018
     /// use syn::{Error, Ident, Result, Token};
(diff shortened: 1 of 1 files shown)


Gankro:
thanks!

konstin:
Thanks!

PR #629, opened 2019-04-30
Tentative beginning to implementing Parse for Expr types

bstrie:
For the various syn::ExprFoo types, some of them implement Parse (e.g. ExprMatch) but most don't. For people writing parsers with syn, it seems to be nice to allow people to write e.g. `let range = syn::parse_macro_input!(item as syn::ExprRange);`, which bypasses the need to parse as Expr and manually match on the variant. The intended use case here is the new feature in Rust 1.34 that allows the argument to proc macros to be any token tree, which will lead to defining trivial parsers for custom syntaxes like `#[foo(1..5, 10..15)]`. The most straightforward way to implement Parse for these types is by simply parsing the input as an Expr and manually returning the data stored in the variant (which is already of the proper type), and returning an Err when encountering an unexpected variant. I wasn't sure if this kind of trivial implementation would be considered "too easy" (as opposed to breaking up the logic in `parse_expr` into each individual type's Parse impl and having `parse_expr` call each impl as needed, but I wasn't sure if this might lead to dangerously deep call trees), so I've only done this for one type (ExprRange) and added a new test. If this looks good then I can do the remainder easily.


src/expr.rs
```diff
@@ -1909,6 +1909,17 @@ pub mod parsing {
         }
     }
 
+    #[cfg(feature = "full")]
+    impl Parse for ExprRange {
+        fn parse(input: ParseStream) -> Result<Self> {
+            let expr: Expr = input.parse()?;
+            match expr {
+                Expr::Range(expr_range) => Ok(expr_range),
+                _ => Err(input.error("expected range expression"))
```
dtolnay: This would incorrectly point to the token after the expression that just got parsed, since that is where `input` would be. It would be better as:

```rust
Err(Error::new_spanned(expr, "..."))
```

dtolnay:
Thanks! I am on board with this approach. If you make a macro to generate these impls, I would consider it sufficient to test only one of the expr types.

bstrie:
Sure, a macro for these impls sounds good to me. I maybe don't need an entire new test and file for testing exactly one of the types, is there an existing test I ought to shove it in instead?

dtolnay:
Let's aspirationally make a test_expr.rs even if this is the one thing tested.

We get most of our test coverage from parsing the complete rust-lang/rust repo in test_round_trip.rs, so the handwritten tests are only for code not covered by that.

bstrie:
Pushed a new commit with the requested changes. Some weirdness: a few of the ExprFoo types (e.g. ExprLit, ExprIf) complained about conflicting Parse impls, and indeed those impls exist, though I don't see corresponding acknowledgement of implementing Parse on the docs pages for these types (see https://docs.rs/syn/0.15.30/syn/struct.ExprWhile.html , and unlike https://docs.rs/syn/0.15.30/syn/struct.ExprMatch.html , which correctly lists ExprMatch as implementing Parse). I've left those lines commented out for now just to designate which types have impls that show up in the code but not in the docs.

One more thing: should all of these impls be tagged with cfg(feature="full")? I noticed that ExprLit (maybe others?) don't gate themselves as such.

bstrie:
It looks like Travis is failing with a particular combination of features enabled ("full" and "parsing"), which I don't quite understand because "full" alone passed and I thought that full implied every other feature?

bstrie:
Oh I see, it was because that test uses --no-default-features and new_spanned is gated behind the printing feature. @dtolnay , should I just use Error::new instead, or should I insert some cfgs in there to make it only use new_spanned when printing is enabled?

bstrie:
It occurs to me that I'm not sure how I would use Error::new, because to get a span at all from an Expr it looks like I need to call .span(), which is blanket impled on ToTokens, which, like new_spanned, is only cfg'd on with the printing feature.

bstrie:
I've cut the Gordian knot in this case by simply gating these new impls on printing as well as full, if only to get the CI passing again. I figure such a thing isn't too bad given that these are just convenience functions. :)

src/expr.rs
```diff
@@ -1909,6 +1909,60 @@ pub mod parsing {
         }
     }
 
+    macro_rules! impl_trivial_parse {
```
dtolnay: I don't think trivial is a useful way to describe these impls. I would recommend something like `impl_by_parsing_expr` for the macro name.

src/expr.rs
```diff
@@ -1909,6 +1909,60 @@ pub mod parsing {
...
+    impl_trivial_parse!(ExprMacro, Macro, "expected macro invocation expression");
+    impl_trivial_parse!(ExprStruct, Struct, "expected struct literal expression");
+    impl_trivial_parse!(ExprRepeat, Repeat, "expected array literal constructed from one repeated element");
+    impl_trivial_parse!(ExprParen, Paren, "expected parenthesized expression");
+    impl_trivial_parse!(ExprGroup, Group, "expected ExprGroup");
+    impl_trivial_parse!(ExprTry, Try, "expected try expression");
+    impl_trivial_parse!(ExprAsync, Async, "expected async block");
+    impl_trivial_parse!(ExprTryBlock, TryBlock, "expected try block");
+    impl_trivial_parse!(ExprYield, Yield, "expected yield expression");
+    impl_trivial_parse!(ExprVerbatim, Verbatim, "expected ExprVerbatim");
```
dtolnay: This should not have a Parse impl.

src/expr.rs
```diff
@@ -1909,6 +1909,60 @@ pub mod parsing {
...
+    impl_trivial_parse!(ExprRange, Range, "expected range expression");
+    impl_trivial_parse!(ExprReference, Reference, "expected referencing operation");
+    impl_trivial_parse!(ExprBreak, Break, "expected break expression");
+    impl_trivial_parse!(ExprContinue, Continue, "expected continue expression");
+    impl_trivial_parse!(ExprReturn, Return, "expected return expression");
+    impl_trivial_parse!(ExprMacro, Macro, "expected macro invocation expression");
+    impl_trivial_parse!(ExprStruct, Struct, "expected struct literal expression");
+    impl_trivial_parse!(ExprRepeat, Repeat, "expected array literal constructed from one repeated element");
+    impl_trivial_parse!(ExprParen, Paren, "expected parenthesized expression");
+    impl_trivial_parse!(ExprGroup, Group, "expected ExprGroup");
```
dtolnay: ExprGroup needs to be handled in the match that you generate for each variant and should not have its own Parse impl. Groups are generated by the macro system to control precedence around metavariables -- imagine them like invisible parens. In your range example, we would want the ExprRange parser to support `#[attr($range, ...)]` where $range is a range expression.

tests/test_parse.rs
```diff
@@ -0,0 +1,16 @@
+extern crate proc_macro2;
```
dtolnay: Please rename to test_expr.rs.

bstrie:
@dtolnay I've made the requested revisions, though I'm not entirely confident that I'm sure of how ExprGroup is supposed to work, largely because I don't know a way to generate one to test with. I've tried parsing "1..10, 2..20" (via both `.parse` and `parse_macro_input!`), but instead of giving me an ExprGroup I just get an error when parsing the comma. Considering that that very example is what I intend to do with syn (having a list of one or more ranges passed in via the attribute), I would be interested to hear more about how you expect this to work; until now I assumed I'd have to make a little parser out of `ExprRange` and `Punctuated` to handle this.

bstrie:
Considering that this also adds test_expr.rs, it might be worth adding in a new test to exercise the ExprGroup arm once I know of a reliable way to generate one.

dtolnay:
Any macro_rules metavariable of type $:expr would turn into ExprGroup. For example:

```rust
extern crate proc_macro;

use proc_macro::TokenStream;
use syn::{parse_macro_input, Expr};

#[proc_macro]
pub fn expr(input: TokenStream) -> TokenStream {
    let expr = parse_macro_input!(input as Expr);
    println!("{:#?}", expr);
    TokenStream::new()
}
```

```rust
macro_rules! m {
    ($e:expr) => {
        expr!($e);
    };
}

m!(1..10);
```

```console
Group(
    ExprGroup {
        attrs: [],
        group_token: Group,
        expr: Range(
            ExprRange {
                ...
```

dtolnay:
A more interesting example:

```rust
macro_rules! m {
    ($e:expr) => {
        expr!($e * 3);
    };
}

m!(1 + 2);
```

```console
Binary(
    ExprBinary {
        attrs: [],
        left: Group(
            ExprGroup {
                attrs: [],
                group_token: Group,
                expr: Binary(
                    /* 1 + 2 */
                ),
            },
        ),
        op: Mul(
            Star,
        ),
        right: Lit(
            /* 3 */
        ),
    },
)
```

dtolnay:
> Considering that that very example is what I intend to do with syn (having a list of one or more ranges passed in via the attribute), I would be interested to hear more about how you expect this to work; until now I assumed I'd have to make a little parser out of `ExprRange` and `Punctuated` to handle this.

I think you would end up needing a Parse impl that calls [`input.parse_terminated(ExprRange::parse)?`](https://docs.rs/syn/0.15.33/syn/parse/struct.ParseBuffer.html#method.parse_terminated).

dtolnay:
Published in 0.15.34.

bstrie:
Thanks! With this PR and your advice my parser is now as simple as the following:

```rust
 let ranges: Punctuated<ExprRange, Token![,]> = input.parse_terminated(ExprRange::parse)?;
 Ok(RangeSequence(ranges.into_iter().collect()))
```

PR #649, opened 2019-06-18
Add `<ParseBuffer as Speculative>::advance_to`, the opposite of `fork`

CAD97:
Revival of #568 now that I've a bit more time and could revisit this. (The syn-dev folder made acquiring the example error messages a lot less involved!) Addresses a large part of #479, as you can now do speculative parsing by forking you input stream, parsing there, and joining back to the fork if parsing was successful.

To clearly show that this is additional functionality rather than core, it is provided as `trait syn::discouraged::Speculative` rather than an inherent impl on `ParseBuffer`, as discussed in #568.


CAD97:
A few thoughts I had after updating this branch:

- Checking that two parse buffers are derived from each other from forking might be doable via checking the cursor scope. I'm not exactly sure how cursors are managed, though, so I'm not confident saying one way or the other. If the cursor `scope` pointer is indeed enough to satisfy the "only a cursor derived from this cell is ever assigned to this cell", then it's definitely better to use it rather than `Rc` ing the buffer `scope: Span` just to track fork derivation.
- `advance_to` currently allows backing up a parse buffer as well as advancing it. I don't know whether this is desirable to keep or prevent, but it can probably be checked by checking which of the buffer's scope cursors is ordered before the other.

dtolnay:
I would love a follow up which uses advance_to to optimize the uses of fork among our existing parsers. Filed #654 to track.

PR #655, opened 2019-06-25
Support `Token![$]` for `token::Dollar` as documented

### LINKED ISSUE #647: Missing `Token![$]` rule for `token::Dollar`
The documentation for `token::Dollar` says one should use `Token![$]` (as all tokens do). The `Token!` macro doesn't have a rule for `$` though, so this doesn't work.

jix:
I added the missing macro rule and changed the single use of `token::Dollar` within syn to use `Token![$]` as the surrounding code used the `Token!` macro for other tokens.

This fixes #647


jix:
Master didn't compile with today's nightly already. It does compile with yesterday's.

codegen/src/parse.rs
```diff
@@ -437,7 +437,7 @@ mod parsing {
                 let content;
                 braced!(content in input);
                 input.parse::<Token![;]>()?;
-                content.parse::<token::Dollar>()?;
+                content.parse::<Token![$]]>()?;
```
dtolnay: This line does not compile because the delimiters are not balanced.
https://travis-ci.org/dtolnay/syn/jobs/550313981

src/token.rs
```diff
@@ -778,6 +778,7 @@ macro_rules! Token {
     (,)           => { $crate::token::Comma };
     (/)           => { $crate::token::Div };
     (/=)          => { $crate::token::DivEq };
+    ($)           => { $crate::token::Dollar };
```
dtolnay: This line does not compile on 1.15.0 which is our minimum supported Rust version. On old compilers we need to not have this rule.
https://travis-ci.org/dtolnay/syn/jobs/550313978

codegen/src/parse.rs
jix: Ah I didn't realize that `cargo test --release --all-features` isn't enough, otherwise I'd have noticed this before opening the PR.

src/token.rs
jix: I'm not sure what the best way is to make this conditional on the compiler version. (I'm also not sure if that is what you're suggesting here.) Also I personally would find it confusing if this feature is dependent on the compiler version used. It'd be easy for code using syn to accidentally depend on a newer rustc version by using `Token![$]`.

Should I open another PR instead that simply documents that `Token![$]` isn't supported and why?

dtolnay: I think having it conditional on the compiler version would be okay. We have some version checks in [syn/build.rs](https://github.com/dtolnay/syn/blob/0.15.38/build.rs) already. You can introduce a cfg that is defined on new enough compilers that are able to support Token![$]. It's possible this will require duplicating the whole Token macro definition but that's fine.

Downstream code accidentally depending on new rustc versions is not a thing we need to consider because regardless of what happens in Syn it's always easy to accidentally depend on new standard library APIs or language features. Crates that care will have a CI setup that catches this (like Syn does).

jix:
I went ahead and did what you suggested (duplicating the macro with a cfg to select the correct version). I bisected this change in rustc down to 1.16 -> 1.17, I made sure `cargo build` succeeds with both of these compiler versions. While making sure that codegen works, I also noticed that the `syn.json` file is generated from that macro, so that is updated too.

PR #676, opened 2019-07-28
Add expr variant for dot-await syntax

### LINKED ISSUE #631: Implement dot-await syntax
We can start by parsing it into the existing syntax tree as ExprField or ExprVerbatim. In the next major version we will want an Expr variant specific to await.

Implementation in rustc: https://github.com/rust-lang/rust/pull/60586

dtolnay:
Closes #631.


dtolnay:
Filed #677 to follow up on the Token macro compatibility with old compilers.

PR #696, opened 2019-08-16
Add Path::get_ident for consumer libraries

infinity0:
I'm trying to update failure-derive to proc-macro2 v1 (and various other newer deps) and this API is needed to easily implement support for the ["_0" "_1" syntax that failure-derive offers](https://github.com/rust-lang-nursery/failure/blob/master/book/src/derive-fail.md#tuple-structs).


dtolnay:
Published in 1.0.3.

PR #706, opened 2019-10-01
Windows downloads rust and runs tests against it

### LINKED ISSUE #572: Tests that need rust-lang/rust checkout don't run on Windows
Specifically, `test_precedence::test_rustc_precedence` and `test_round_trip::test_round_trip` both call `common::clone_rust` which runs `tests/clone.sh`, which obviously doesn't work on Windows.

There are several ways this could be addressed:

- Detect Windows and fail the test with a more useful error than ``called `Result::unwrap()` on an `Err` value: Os { code: 193, kind: Other, message: "%1 is not a valid Win32 application." }``
- Detect Windows and ignore the tests (`#[cfg_attr(target_os = "windows", ignore)]`)
- Rewrite `clone.sh` using Rust
- Provide a Powershell equivalent of `clone.sh` and use that on Windows

AndyGauge:
Fixes #572


AndyGauge:
AppVeyor doesn't run the tests.  Is that by design?

AndyGauge:
all the files that were added had permission bits changed.  That's what I get for using windows lol.  Let me fix it.

AndyGauge:
That's a great solution.  I had the same thought about an hour ago, after submitting the PR.

PR #707, opened 2019-10-02
Rust implementation of download and extract source code

### LINKED ISSUE #572: Tests that need rust-lang/rust checkout don't run on Windows

AndyGauge:
Fixes #572


tests/repo/mod.rs
```diff
@@ -52,7 +55,36 @@ pub fn base_dir_filter(entry: &DirEntry) -> bool {
...
+                }
+            }
+            let root_folder = zip.by_index(0).unwrap().sanitized_name();
+            let rust_path = Path::new("tests/rust");
+            if rust_path.exists() {
+                remove_dir_all(rust_path)?;
+            }
+            rename(root_folder, rust_path)?;
+            if let Ok(mut commit) = File::create(&Path::new("tests/rust/COMMIT")) {
+                commit.write_all(REVISION.as_bytes())?;
```
dtolnay: It would be simpler to use https://doc.rust-lang.org/std/fs/fn.write.html:

```rust
fs::write("tests/rust/COMMIT", REVISION)?;
```

tests/repo/mod.rs
```diff
@@ -52,7 +55,36 @@ pub fn base_dir_filter(entry: &DirEntry) -> bool {
...
-pub fn clone_rust() {
-    let result = Command::new("tests/clone.sh").status().unwrap();
-    assert!(result.success());
+pub fn clone_rust() -> io::Result<()> {
+    if match read_to_string("tests/rust/COMMIT") {
+        Err(_) => true,
+        Ok(contents) => contents.trim() != REVISION,
+    } {
+        let url = format!("https://github.com/rust-lang/rust/archive/{}.zip", REVISION);
+        let mut tmpfile = tempfile::tempfile().unwrap();
```
dtolnay: Would it be possible to do this without saving the file to disk? These files are reasonably small.

dtolnay:
Ping @AndyGauge &mdash; are you still planning to work more on this?

AndyGauge:
Yes, I’ll see it through.  I’ve been overcommitted on other projects recently.

Maybe 2 weeks from now or so.

tests/repo/mod.rs
```diff
@@ -52,7 +55,36 @@ pub fn base_dir_filter(entry: &DirEntry) -> bool {
...
 
-pub fn clone_rust() {
-    let result = Command::new("tests/clone.sh").status().unwrap();
-    assert!(result.success());
+pub fn clone_rust() -> io::Result<()> {
+    if match read_to_string("tests/rust/COMMIT") {
+        Err(_) => true,
+        Ok(contents) => contents.trim() != REVISION,
+    } {
+        let url = format!("https://github.com/rust-lang/rust/archive/{}.zip", REVISION);
```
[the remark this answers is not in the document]
AndyGauge: I incorrectly assumed tar.gz was not supported on windows, I just proved that wrong and am rewriting the PR to use tar.gz and flate crate

tests/repo/mod.rs
AndyGauge: I will check on that.

AndyGauge: I haven't found a way other than memmap, which I'm not comfortable adding that much complexity.

tests/repo/mod.rs
```diff
@@ -52,7 +55,36 @@ pub fn base_dir_filter(entry: &DirEntry) -> bool {
...
+        if let Ok(_request) = reqwest::get(url.as_str()).unwrap().copy_to(&mut tmpfile){
+            let mut zip = zip::ZipArchive::new(tmpfile).unwrap();
+            for i in 0..zip.len() {
+                let mut file = zip.by_index(i).unwrap();
+                let outpath = file.sanitized_name();
+                if file.name().ends_with('/') {
+                    create_dir_all(&outpath).unwrap()
+                }
+                else {
+                    let mut outfile = File::create(&outpath).unwrap();
```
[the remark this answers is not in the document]
AndyGauge: Most of this complexity will be removed by using flate2.

tests/repo/mod.rs
```diff
@@ -52,7 +55,36 @@ pub fn base_dir_filter(entry: &DirEntry) -> bool {
...
-    let result = Command::new("tests/clone.sh").status().unwrap();
-    assert!(result.success());
+pub fn clone_rust() -> io::Result<()> {
+    if match read_to_string("tests/rust/COMMIT") {
+        Err(_) => true,
+        Ok(contents) => contents.trim() != REVISION,
+    } {
+        let url = format!("https://github.com/rust-lang/rust/archive/{}.zip", REVISION);
+        let mut tmpfile = tempfile::tempfile().unwrap();
+        if let Ok(_request) = reqwest::get(url.as_str()).unwrap().copy_to(&mut tmpfile){
```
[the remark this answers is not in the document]
AndyGauge: I'm rewriting the return type to Box<dyn std::error::Error>.  I'm `.expect`ing the result anyway so it shouldn't have a major impact, but at least the errors will make it to the caller.

dtolnay:
Thanks -- I removed the tempfile stuff in 02c6857e1bf2df78cd1e9f25fbb1d5fcfa3fdce0, we don't need the tar.gz written to the filesystem or memmap.

PR #712, opened 2019-10-26
Add `len` and `is_empty` methods to `Fields`

mjbshaw:
There are a few situations where I've been using `fields.iter().len()`. This works, but having inherent `len()` and `is_empty()` methods would make things feel more idiomatic.


Kampfkarren:
Why was this put into a patch release and not in a minor release if it's a new API? 😕

PR #716, opened 2019-11-08
Parse TypeTuple with one element now requires trailing comma

8BitMate:
Fixes bug where a single type surrounded by parentheses could be parsed as TypeTuple. It now requires a trailing comma when there is only one type.


dtolnay:
Published in syn 1.0.8.

Issue #720, opened 2019-11-23
Visibility parse failure due to speculative behaviour


mystor:
This valid rust code fails to parse with `syn` ([playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=db75eb70c47e972591d7edb4d390d09a)):

```rust
pub struct A(pub (crate::B, crate::C));
```

## Explanation

This parsing bug is caused by the `(crate ...)` block being speculatively parsed as the `(crate)` part of `pub(crate)`. As the `parenthesized!` `ParseBuffer` is not parsed to completion, it sets the shared `unexpected` error when it's dropped, but does not directly return an error.

This works OK in the non-speculative case, as the overall parse will fail if any grouped sections have failed to parse correctly. Unfortunately, it does not interact well with speculatively parsing groups. The speculative parse appears to succeed, as no error is directly returned, which leads to the wrong branch being taken by the `Visibility` parser.

https://github.com/dtolnay/syn/blob/339336cbb73c5a3413c1ebd213a8794fca7e1fb8/src/data.rs#L330-L354

## Potential Fixes

The `Visibility` parser is fairly simple, so can be fixed by explicitly checking for the end of the subgroup while speculating. This doesn't expand for users speculatively parsing more complex types, however. 

It may be necessary to publicly expose an API for checking if a `ParseStream` is currently in an error state due to unexpected tokens (like [`check_unexpected`](https://github.com/dtolnay/syn/blob/339336cbb73c5a3413c1ebd213a8794fca7e1fb8/src/parse.rs#L965-L970)). This could be used, along with careful scoping of group parsers, to accurately check for speculative parsing success. This workaround would also need to documented as a caveat by `ParseStream::fork`.

Issue #721, opened 2019-11-23
Speculative::advance_to discards `unexpected`


mystor:
The current implementation of `advance_to` discards any current or future `unexpected` spans from the speculative parsing branch, meaning that bad parses can appear to succeed. ([playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=479ac297014237dcf8a9663c4e08b263))

The first solution I thought of was to check `unexpected` on the fork in [`advance_to`](https://github.com/dtolnay/syn/blob/339336cbb73c5a3413c1ebd213a8794fca7e1fb8/src/discouraged.rs#L162-L170), and copy it over if it has been set. This will fix well-scoped speculative parses (such as `issue1`), but does not handle subparsers which have yet to be dropped (such as `issue2`).

It's not the sleekest fix, but a potential workaround to the other set of issues would be to change the type of the [`unexpected` field](https://github.com/dtolnay/syn/blob/339336cbb73c5a3413c1ebd213a8794fca7e1fb8/src/parse.rs#L264) to something like this:

```rust
enum Unexpected {
    None,
    Span(Span),
    Joined(Rc<Cell<Unexpected>>),
}

struct ParseBuffer<'a> {
    // ...
    unexpected: Rc<Cell<Unexpected>>,
}
```

The `advance_to` method would then change the state of the speculative parser's `unexpected` field to `Joined(original.unexpected.clone())`. This would create a chain of `Rc`s leading up to the real parser, so that `Unexpected` errors which appear after `advance_to` is called get forwarded to the right place.

(cc #720, #654)

PR #722, opened 2019-11-26
Closure-based speculative parsing API

mystor:
This is a (very early draft of a) potential fix for the speculative parsing issues in #720 and #721. It introduces a new API for speculative parsing. The `speculate` method takes a `FnOnce(ParseStream) -> Result<T>`, and returns an `Option<T>`, only advancing the `ParseStream` if the inner parse succeeds.

The primary difference between this method and `fork`/`advance_to` is the interaction with the `unexpected` attribute. The forked `ParseStream` is checked for `unexpected` errors before determining whether the speculative parse succeeded. The introduction of a closure scope helps ensure that nested `ParseBuffer`s have been dropped before the check occurs, which helps with issue 2 in #721. 

I also introduce a `speculate_peek` method (all names super bikesheddable) which also checks `unexpected` errors, but does not `advance_to` if the inner parse succeeds.

## Example

A number of `fork()` callers have been changed in the 2nd commit, including the `Visibility::parse_pub` method which caused #720. The implementation now looks as follows:

```rust
let pub_token = input.parse::<Token![pub]>()?;

match input.speculate(|ahead| {
    let content;
    let paren_token = parenthesized!(content in ahead);

    if content.peek(Token![crate])
        || content.peek(Token![self])
        || content.peek(Token![super])
    {
        Ok(VisRestricted {
            pub_token,
            paren_token,
            in_token: None,
            path: Box::new(Path::from(content.call(Ident::parse_any)?)),
        })
    } else {
        Ok(VisRestricted {
            pub_token,
            paren_token,
            in_token: Some(content.parse()?),
            path: Box::new(content.call(Path::parse_mod_style)?),
        })
    }
}) {
    Some(restricted) => Ok(Visibility::Restricted(restricted)),
    None => Ok(Visibility::Public(VisPublic { pub_token })),
}
```

## Known Problems

* The function boundary can encourage too much to be parsed speculatively. For example, in the previous "example" code, the `path` field in `pub(in FOO)` should not be parsed speculatively, however performing the parse outside of the `speculate` call would be much more verbose.
   
   It may be possible to add a `ahead.nonspeculative(|ahead| ...)` helper if the input & return types of the `speculate`  are custom wrapper types. Alternatively, a `ahead.commit()` (`ahead.stop_speculating()`?) method could also be used to halt speculation, and make errors produced after that point trigger a "real" error rather than a speculation one by setting a flag on the forked ParseBuffer. (The `commit()` would have to be ignored if an `unexpected` flag had already been set to ensure the parse recovers in that scenario)
   
   Either of these approaches would require changing the return type to `Result<Option<T>>`, which should be OK.

* It's fairly common to perform a series of `peek` checks while speculating, and then return a dummy error (which will be ignored) if they fail. It might be worthwhile to use a different error type so these are easier to write & read.

* Documentation needs a lot of updating. The current docs are largely copy-pasted from `advance_to`.

## Breaking Changes

I wanted to add this new method to the `parse::discouraged::Speculative` extension trait, as it fills the same role as the existing `advance_to` method. That extension trait is not `Sealed`, meaning it is _technically_ possible for downstream crates to implement it for local types.

In this patch stack I change this trait to be `Sealed`, and add new methods to it. This is technically a breaking change.

### DIFF
--- a/src/data.rs
+++ b/src/data.rs
@@ -238,6 +238,7 @@ pub mod parsing {
     use super::*;
 
     use crate::ext::IdentExt;
+    use crate::parse::discouraged::Speculative;
     use crate::parse::{Parse, ParseStream, Result};
 
     impl Parse for Variant {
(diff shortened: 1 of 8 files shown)


mystor:
I didn't realize that `token::*` doesn't implement `Copy` unless the `clone-impls` feature has been enabled. I suppose I'll need to remove the dependency on `pub_token` being `Copy` at some point.

It also appears that some nightly changes broke the rustc tests (I think)

mystor:
The `Rc`-chain option fixes the issue with `advance_to` discarding the error, but doesn't provide a way for non-internal code to query the error state of the `unexpected` value. To fix #720, which doesn't use `advance_to` yet, we'd need to either expose `ParseBuffer::check_unexpected` or an API which uses it, like this one, as well.

dtolnay:
For #720 I think I'd like Visibility::parse_pub to do its own is_empty check. This is a quirk of visibility syntax. It's speculating into the parentheses and needs to encode the fact that `pub ( crate )` vs `pub ( crate ... )` makes a difference.

mystor:
Closing in favour of #723.

PR #723, opened 2019-11-28
Fix Visibility parsing errors, and optimize with 'advance_to'

### LINKED ISSUE #720: Visibility parse failure due to speculative behaviour

mystor:
As suggested in https://github.com/dtolnay/syn/pull/722#issuecomment-558451027, the Visibility patch (patch 2) does not introduce new APIs for observing 'unexpected' errors, and instead manually checks for 'content.is_empty()' when speculatively parsing in `Visibility`.

The final patch fixes 'unexpected' error propagation when using 'advance_to'. This ended up being more complex than I expected, as the fork's root parser is likely to set `unexpected` when it is dropped, which should not cause an error in the advanced stream. This was worked around using `Cell` directly within `ParseBuffer`, in addition to within `Rc`s. 

Fixes #720, #721

### DIFF
--- a/src/data.rs
+++ b/src/data.rs
@@ -238,6 +238,7 @@ pub mod parsing {
     use super::*;
 
     use crate::ext::IdentExt;
+    use crate::parse::discouraged::Speculative;
     use crate::parse::{Parse, ParseStream, Result};
 
     impl Parse for Variant {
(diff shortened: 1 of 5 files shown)


dtolnay:
Published in 1.0.9.

PR #744, opened 2020-01-05
Make the group::parse_group public (and others)

pksunkara:
I am working on building a substitute `Expr` enum which only allows a subset of Rust expressions. I was able to finish off it except for this one small thing that I needed to use.

This small function contains a lot of code I don't want to duplication since it doesn't depend on `Expr`. If you make this `pub`, my use case would be satisfied.

Can you please merge and release a new version ASAP? Thanks.

### DIFF
--- a/src/expr.rs
+++ b/src/expr.rs
@@ -2661,9 +2661,14 @@ pub(crate) mod parsing {
         }
     }
 
+
+    /// Determines whether this is a named member or not
+    ///
+    /// *This function is available if Syn is built with the `"full"`
+    /// feature.*
     #[cfg(feature = "full")]
     impl Member {
-        fn is_named(&self) -> bool {
+        pub fn is_named(&self) -> bool {
             match *self {
                 Member::Named(_) => true,
                 Member::Unnamed(_) => false,
--- a/src/group.rs
+++ b/src/group.rs
@@ -60,8 +60,10 @@ pub fn parse_brackets<'a>(input: &ParseBuffer<'a>) -> Result<Brackets<'a>> {
     })
 }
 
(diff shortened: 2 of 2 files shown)


pksunkara:
Made `Member::is_named()` public. I know it's a simple `match`, but it's good to have this API in the struct itself.

dtolnay:
APIs that are marked #\[doc(hidden)\] and commented with `// Not public API` are not supposed to be used outside of Syn. Your code will break when we change them.

dtolnay:
Here is a parser for invisible groups using only the public API.

```rust
use proc_macro2::{Delimiter, TokenStream, TokenTree};
use std::iter::FromIterator;
use syn::parse::{Parse, ParseStream, Result};
use syn::token::{self, Group};
use syn::{parse_quote, Expr};

#[derive(Debug)]
struct ExprGroup {
    group_token: Group,
    expr: Box<Expr>,
}

impl Parse for ExprGroup {
        let group = match input.parse::<TokenTree>()? {
            TokenTree::Group(group) => group,
            _ => return Err(input.error("expected invisible group")),
        };
        let group_token = token::Group(group.span());
        let expr = syn::parse2(group.stream())?;
        Ok(ExprGroup { group_token, expr })
    }
}

fn main() {
    let tokens = TokenStream::from_iter(vec![TokenTree::Group(proc_macro2::Group::new(
        Delimiter::None,
        parse_quote!(1 + 1),
    ))]);
    let e = syn::parse2::<Expr>(tokens).unwrap();
    println!("{:#?}", e);
}
```

pksunkara:
That worked. Thanks.

PR #747, opened 2020-01-15
Add Error::set_span

CreepySkeleton:
This function is quite usefult for patching error messages, like
```rust
let res: Visibility = syn::parse_str("...").map_err(|e| {
    // the current span is call_site(), but we have the more fitting place to point to 
    e.set_span(some);
    e
})?;

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -188,6 +188,14 @@ impl Error {
         start.join(end).unwrap_or(start)
     }
 
+    /// Set the source location of the error for all contained messages.
+    pub fn set_span(&mut self, span: Span) {
+        for message in self.messages.iter_mut() {
+            message.start_span = ThreadBound::new(span);
+            message.end_span = ThreadBound::new(span);
+        }
+    }
+
     /// Render the error as an invocation of [`compile_error!`].
     ///
     /// The [`parse_macro_input!`] macro provides a convenient way to invoke

### CODE REVIEW

**Comment by CreepySkeleton** [comment]
The first one is https://github.com/TeXitoi/structopt/blob/d804cecd7d5a4a78038919580e71f2f3cde4230a/structopt-derive/src/parse.rs#L293..L302 . We don't have the needed span at the point the error originates so we pass it up to the upper level and patch it there.

The second one is in [this PR](https://github.com/Hoverbear/getset/pull/48/files#diff-6755e3579c3e8158c7d672a1ddebb24dR42)
```rust
.map(|v| {
    syn::parse_str(v)
        .unwrap_or_else(|_| abort!(s.span(), "invalid visibility found"))
})
```
I would like it to look like 
```rust
.map(|v| {
    syn::parse_str(v)
        .map_err(|e| { e.set_span(s.span()); e })
        .expect_or_abort("invalid visibility found")
})
```
We can't use `LitStr::parse` there since we're parsing a subslice, not the whole literal. Building a new `LitStr` would be even uglier.

There are few more snippets I could show, but they are very similar to the two above.

dtolnay:
Your first use case is currently written as:

```rust
let attrs: StructOptAttributes = parse2(attr.tokens.clone())
    .map_err(|e| match &*e.to_string() {
        // this error message is misleading and points to Span::call_site()
        // so we patch it with something meaningful
        "unexpected end of input, expected parentheses" => {
            let span = attr.path.span();
            let patch_msg = "expected parentheses after `structopt`";
            syn::Error::new(span, patch_msg)
        }
        _ => e,
    })
    .unwrap_or_abort();
attrs.attrs
```

I would prefer to see this either as:

```rust
if attr.tokens.is_empty() {
    let span = attr.path.span();
    let patch_msg = "expected parentheses after `structopt`";
    abort!(span, patch_msg);
} else {
    parse2::<StructOptAttributes>(attr.tokens.clone())
        .unwrap_or_abort()
        .attrs
}
```

or:

```rust
attr.parse_args_with(Punctuated::<StructOptAttr, Token![,]>::parse_terminated)
    .unwrap_or_abort()
```

For the second use case:

```rust
.map_err(|e| { e.set_span(s.span()); e })
```

I would expect:

```rust
.map_err(|e| Error::new(s.span(), e))
```

or for similar use cases where the error would need to span multiple tokens:

```rust
.map_err(|e| Error::new_spanned(s, e))
```

CreepySkeleton:
Well, that settles it. You are totally right about the second case.

It's a pity that the documentation for `parse_args_with` is deficient, in particular it doesn't say anything about what happens when no `(`/`{`/`[` groups are present. Would you be willing to accept a PR fixing this?

PR #798, opened 2020-05-16
Make `cargo test` fail more helpfully

dtolnay:
<p align="center"></p>

<p align="center"></p>

<p align="center"></p>


dtolnay:
cc https://github.com/rust-lang/cargo/issues/1106 + https://github.com/rust-lang/cargo/issues/985 -- for *some* things that a build script would normally do, a proc macro can do it instead without the output being suppressed.

PR #801, opened 2020-05-16
Update doc comments about feature availability.

### LINKED ISSUE #796: Improve documentation of item availability depending on feature flags.
I've been writing a procedural macro recently (and thinking about how much harder it would be without `syn` – this crate is awesome!). Most of the documentation contains comments about the availability of features which look a little like this:
Screenshot 2020-05-15 at 18 43 55

`"This type is available if Syn is built with the "derive" or "full" feature."`

I find that these would be clearer if they said something like `This type is available only if Syn is built with the "derive" or "full" feature.` or `This type is available iff Syn is built with the "derive" or "full" feature.` 

I'd be happy to do this (with find and replace or changing doc comments about availability of features to be generated through a macro).

teymour-aldridge:
This fixes #796

### DIFF
--- a/codegen/src/fold.rs
+++ b/codegen/src/fold.rs
@@ -272,7 +272,7 @@ pub fn generate(defs: &Definitions) -> Result<()> {
             ///
             /// [module documentation]: self
             ///
-            /// *This trait is available if Syn is built with the `"fold"` feature.*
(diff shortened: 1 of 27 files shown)


teymour-aldridge:
You’re welcome.

PR #813, opened 2020-05-21
Parse const trait impls

### LINKED ISSUE #766: Parse const trait impls
As seen in https://github.com/rust-lang/rust/blob/8aa9d2014f4e5258f83b907e8431c59a33acdae7/src/test/ui/rfc-2632-const-trait-impl/const-and-non-const-impl.rs#L6.

```rust
impl const std::ops::Add for i32 {
    type Output = Self;

    fn add(self, rhs: Self) -> Self {
        self + rhs
    }
}
```

taiki-e:
It cannot be parsed as `ItemImpl`, so it parsed as `Item::Verbatim`.

Closes #766

### DIFF
--- a/src/item.rs
+++ b/src/item.rs
@@ -1266,7 +1266,11 @@ pub mod parsing {
                 {
                     input.parse().map(Item::Trait)
                 } else if lookahead.peek(Token![impl]) {
-                    input.parse().map(Item::Impl)
+                    if let Some(item) = parse_impl(input, true)? {
+                        Ok(Item::Impl(item))
+                    } else {
+                        Ok(Item::Verbatim(verbatim::between(begin, input)))
+                    }
                 } else {
                     Err(lookahead.error())
                 }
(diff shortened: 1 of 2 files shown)


src/item.rs
```diff
@@ -2461,45 +2469,61 @@ pub mod parsing {
...
 
-            let content;
-            let brace_token = braced!(content in input);
-            let inner_attrs = content.call(Attribute::parse_inner)?;
+        let is_const_impl = allow_const_impl
+            && (input.peek(Token![const]) || input.peek(Token![?]) && input.peek2(Token![const]));
+        if is_const_impl {
+            input.parse::<Option<Token![?]>>()?;
+            input.parse::<Token![const]>()?;
+        }
```
taiki-e, line 2500: This accepts inherent const impl because rustc [syntactically accepts](https://play.rust-lang.org/?version=nightly&mode=debug&edition=2018&gist=587044e73f6559fd34953fd4c06d22c9) it (but [not semantically](https://github.com/rust-lang/rust/blob/master/src/test/ui/rfc-2632-const-trait-impl/inherent-impl.rs)).

Issue #264, opened 2017-12-05
Parse const generic argument as ExprVerbatim when building without feature(full)


dtolnay:
Use the verbatim variant idea from #251 to parse a braced expression in #263 as just a verbatim tokenstream expression.

mystor:
I think it might be better to add support for the minimum set of variants required to parse { EXPR }. I can imagine people using custom derive wanting to be able to interpret `Foo<A, { X + 1 }>`

It'd be good to know how much new stuff we'd have to pull in to enable that.

dtolnay:
I guess I have the opposite perspective. Typically in custom derives you just need the field types to be shuffled from the input to the right places in the output. In that sense the following field types are basically equivalent and you wouldn't need to look inside the tokens of either.

```rust
#[derive(Serialize)]
struct Mystor {
    g: Foo<{X + 1}>,
    m: Foo!{X + 1},
}
```

But I just remembered we already parse a subset of expressions in array lengths, like `[u32; X.bits]`. That same subset could work here with everything else being verbatim tokens.

mystor:
You make a good point. I'm tempted now to see if we can improve build performance in non-full builds by not building any Expr variants other than `ExprVerbatim` and perhaps `Lit`.

dtolnay:
Triage: I don't think we need to care about this until const generics are implemented in nightly.

dtolnay:
Triage: still not available yet in nightly.

DutchGhost:
What is the update on this?

I tried to parse this:
```Rust
#[validate2]
impl<const S: &'static str> Example<{ S }> {
    pub fn new(n: usize) -> Self {
        Self { n }
    }
}
```

with the following code:
```Rust
#[proc_macro_attribute]
pub fn validate2(_attr: TokenStream, item: TokenStream) -> TokenStream {
    let mut item_impl = parse_macro_input!(item as syn::ItemImpl);
    panic!("GOT HERE");
    quote!(item_impl).into()
}
```

However, I get an error message:
```
error: expected one of: `for`, parentheses, `fn`, `unsafe`, `extern`, identifier, `::`, `<`, square brackets, `*`, `&`, `!`, `impl`, `_`, lifetime
  --> examples\string.rs:11:6
   |
11 | impl<const S: &'static str> Example<{ S }> {
   |      ^^^^^
```

I understand that const generics are on nightly, and the feature isn't complete yet, but on the other hand, simple const generics are usable, and I would like to be able to use syn with them.

[...]

PR #816, opened 2020-05-24
Parse const generics on methods with feature(full)

yodaldevoid:
I must admit I have not extensively tested this, but the code is simple enough and taken from the support for const generics on types so I have a lot of confidence in this.


yodaldevoid:
@DutchGhost This should fix the problem you were talking about on #264 if you are using feature(full).

PR #858, opened 2020-06-24
Type unit

sethp:
Hello! I am new to rust, so please let me know if there is a different answer to my question. I am trying to build a match with an arm for functions with no return signature or functions that return unit, and was finding it somewhat difficult to do so. Previously, `()` in a type position would be parsed as `Type::Tuple` with empty `elems`, like so:

```
Tuple(TypeTuple { paren_token: Paren, elems: [] })
```

But with this change I believe I can express my pattern like this (with `#![feature(box_patterns)]`):

```
match ret {
    ReturnType::Default | ReturnType::Type(_, box Type::Unit(..)) => ...
}
```

Is it worth it? Well, that's for you to decide :)

### DIFF
--- a/src/ty.rs
+++ b/src/ty.rs
@@ -67,7 +67,7 @@ ast_enum_of_structs! {
         Tuple(TypeTuple),
 
         /// The unit type: `()`
-        Unit,
+        Unit(TypeUnit),
 
         /// Tokens in type position not interpreted by Syn.
         Verbatim(TokenStream),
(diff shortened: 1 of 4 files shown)


sethp:
Fair enough! I had thought of tuples as having length at least 1, and "unit" as something distinct, but thinking of unit as a nullary tuple makes a lot of sense too.

Either way, I had fun working on the patch, and I feel like I learned a lot – you've got a pretty neat project here. And thank you for the suggestion! That looks like it'll work really well for me.

PR #880, opened 2020-08-04
Accept traits with parenthesized path arguments in impls

alecmocatta:
This lets `ItemImpl`s, i.e. `impl Trait for Type {}`, include parenthesized path arguments on `Trait`. An example would be `impl FnOnce() for X {}`.

This is in line with `rustc` which parses successfully, but bails later with "associated type bindings are not allowed here".


dtolnay:
Published in 1.0.38.

alecmocatta:
Brilliant, thanks for the swift release @dtolnay!

PR #891, opened 2020-08-31
Avoid panic on printing incomplete qualified path

### LINKED ISSUE #882: 'attempt to subtract with overflow' in Path::span
Sometimes calling `span` on a `Path` will panic with the error `attempt to subtract with overflow`.  This is very difficult to debug, since backtraces don't work in proc macros.  Here is a test case that reproduces the problem:
```rust
#[test]
fn subtract_with_overflow() {
    use quote::quote;
    use syn::spanned::Spanned;
    use syn::Type;

    let mut t: Type = syn::parse2(quote!(<Self as A>::Q)).unwrap();
    if let Type::Path(ref mut tp) = &mut t {
        tp.path.segments.pop();
        tp.path.segments.pop();
    }
    dbg!(t.span());
}
```
```
$ cargo test --all-features --test test_path -- --nocapture subtract_with_overflow

   Compiling syn v1.0.38 (/usr/home/somers/src/rust/syn)

NOTE:  use --release
  Syn's test suite has some tests that run on every source file
  and test case in the rust-lang/rust repo, which can be pretty
  slow in debug mode. Consider running cargo test with `--release`
  to speed things up.

    Finished test [unoptimized + debuginfo] target(s) in 3.98s
     Running target/debug/deps/test_path-12ef20a528c1cde0

running 1 test
thread 'subtract_with_overflow' panicked at 'attempt to subtract with overflow', src/path.rs:715:17
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
test subtract_with_overflow ... FAILED

failures:

failures:
    subtract_with_overflow

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 1 filtered out

error: test failed, to rerun pass '--test test_path'
```

taiki-e:
Fixes #882

### DIFF
--- a/tests/test_path.rs
+++ b/tests/test_path.rs
@@ -50,3 +50,58 @@ fn parse_interpolated_leading_component() {
     }
     "###);
 }
+
+#[test]
+fn print_incomplete_qpath() {
+    use quote::ToTokens;
+    use syn::TypePath;
+
+    // qpath with `as` token
+    let mut ty: TypePath = syn::parse_quote!(<Self as A>::Q);
+    snapshot!(ty.to_token_stream(), @r###"
+    TokenStream(`< Self as A > :: Q`)
+    "###);
+    assert!(ty.path.segments.pop().is_some());
+    snapshot!(ty.to_token_stream(), @r###"
+    TokenStream(`< Self as A > ::`)
+    "###);
+    assert!(ty.path.segments.pop().is_some());
+    snapshot!(ty.to_token_stream(), @r###"
+    TokenStream(`< Self >`)
+    "###);
+    assert!(ty.path.segments.pop().is_none());
+
+    // qpath without `as` token
+    let mut ty: TypePath = syn::parse_quote!(<Self>::A::B);
+    snapshot!(ty.to_token_stream(), @r###"
+    TokenStream(`< Self > :: A :: B`)
+    "###);
+    assert!(ty.path.segments.pop().is_some());
+    snapshot!(ty.to_token_stream(), @r###"
+    TokenStream(`< Self > :: A ::`)
+    "###);
+    assert!(ty.path.segments.pop().is_some());
+    snapshot!(ty.to_token_stream(), @r###"
+    TokenStream(`< Self > ::`)
+    "###);
+    assert!(ty.path.segments.pop().is_none());
+
+    // normal path
+    let mut ty: TypePath = syn::parse_quote!(Self::A::B);
+    snapshot!(ty.to_token_stream(), @r###"
+    TokenStream(`Self :: A :: B`)
+    "###);
+    assert!(ty.path.segments.pop().is_some());
+    snapshot!(ty.to_token_stream(), @r###"
+    TokenStream(`Self :: A ::`)
+    "###);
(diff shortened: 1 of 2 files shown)


tests/test_path.rs
```diff
@@ -50,3 +50,17 @@ fn parse_interpolated_leading_component() {
...
+    use quote::ToTokens;
+    use syn::Type;
+
+    let mut ty: Type = syn::parse_quote!(<Self as A>::Q);
+    if let Type::Path(tp) = &mut ty {
+        tp.path.segments.pop();
+        tp.to_token_stream();
+        tp.path.segments.pop();
+    }
+    ty.to_token_stream();
```
taiki-e, line 65: btw, the output of first `.to_token_stream()` is currently broken (second is printed as `<Self>`)
: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=e063b5e4bb31119e2bc57cae3d7cdc22

I think it might be preferable to print it like `<Self as A>` (because we can write like `quote!(#tp::#path_assoc_item)`), but I haven't looked into its effect on others, so I don't touch it in this PR.

tests/test_path.rs
```diff
@@ -50,3 +50,17 @@ fn parse_interpolated_leading_component() {
...
     "###);
 }
+
+#[test]
+fn print_incomplete_qpath() {
+    use quote::ToTokens;
+    use syn::Type;
+
+    let mut ty: Type = syn::parse_quote!(<Self as A>::Q);
+    if let Type::Path(tp) = &mut ty {
```
dtolnay, line 60: Recommend parsing directly to TypePath to bypass the unsuccessful branch of this `if let`.

```suggestion
    let mut ty: TypePath = syn::parse_quote!(<Self as A>::Q);
```

tests/test_path.rs
dtolnay: Please update both of these to_token_stream calls to compare the return value against what is expected, and update the ToTokens impl to produce something reasonably useful for the cases that we care about supporting.

tests/test_path.rs
```diff
@@ -50,3 +50,58 @@ fn parse_interpolated_leading_component() {
...
+    snapshot!(ty.to_token_stream(), @r###"
+    TokenStream(`< Self as A > :: Q`)
+    "###);
+    assert!(ty.path.segments.pop().is_some());
+    snapshot!(ty.to_token_stream(), @r###"
+    TokenStream(`< Self as A > ::`)
+    "###);
+    assert!(ty.path.segments.pop().is_some());
+    snapshot!(ty.to_token_stream(), @r###"
+    TokenStream(`< Self >`)
```
taiki-e, line 70: It doesn't print `::` unlike the other cases, but it is probably okay as the user may have manually set `QSelf::leading_colon` to `None`.

taiki-e:
@dtolnay: Addressed your review and added tests.

tests/test_path.rs
taiki-e: Hmm, `QSelf` itself is basically in inside `Option`, so the chance of manually setting `QSelf::leading_colon` to `None` may be very low.

dtolnay (approved):
Thanks!

PR #895, opened 2020-09-06
filter for float finiteness in proc_macro2 literal construction

sameer:
Ran into another panic while fuzzing:

```rust
fn a() {
    X.-0.9E999999
}
```
panics with `thread 'main' panicked at 'assertion failed: f.is_finite()', /home/purisame/.cargo/registry/src/github.com-1ecc6299db9ec823/proc-macro2-1.0.19/src/lib.rs:1073:9`

but:

```rust
fn a() {
    -0.9E999999
}
```

does not. Syn is calling a function that panics for non-finite floats when parsing expressions, so I've filtered those out.

### Parsing non-finite floats

This also raises some other questions -- why does parsing a non-finite float work in the first place?

```
                                        expr: Lit(
                                            ExprLit {
                                                attrs: [],
                                                lit: Float(
                                                    LitFloat {
                                                        token: 0.9E999999,
                                                    },
                                                ),
                                            },
                                        ),
```

If we ever have f128, f80, etc. in Rust I think this behavior is fine. In which case, should the `other_suffix` case should fall back to

```
            let stream = repr.parse::<TokenStream>().unwrap();
            match stream.into_iter().next().unwrap() {
                TokenTree::Literal(l) => Some(l),
                _ => unreachable!(),
            }
```

if parsing returns an error?

### DIFF
--- a/src/lit.rs
+++ b/src/lit.rs
@@ -103,7 +103,7 @@ struct LitIntRepr {
 ast_struct! {
     /// A floating point literal: `1f64` or `1.0e10f64`.
     ///
-    /// Must be finite. May not be infinte or NaN.
+    /// Must be finite. May not be infinite or NaN.
     pub struct LitFloat {
         repr: Box<LitFloatRepr>,
     }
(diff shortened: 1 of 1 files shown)


dtolnay (approved):

dtolnay:
> why does parsing a non-finite float work?

0.9E999999 isn't non-finite. Someone making an arbitrary precision macro like `apfloat!(0.9E999999)` would need the value.

> other_suffix case should fall back to

TokenStream::from_str can't make negative literal tokens.

sameer:
Ah ok, that makes sense. Thanks for explaining! :+1:

PR #896, opened 2020-09-14
Add case to parse_macro_input! to use a Parser function

sbrocket:
This would have been useful in https://github.com/dtolnay/cxx/pull/298


dtolnay (approved):
Thanks!

dtolnay:
Published in 1.0.41.

PR #922, opened 2020-11-08
Document completion check on syn::parse2

Tamschi:
The free-standing `parse2`'s documentation doesn't mention that it can error even if the underlying `Parse::parse` succeeds.  
[This tripped me up a bit.](https://github.com/Tamschi/call2-for-syn/issues/1)

This pull request copies the relevant phrase from `Parser::parse2`'s documentation.


dtolnay (approved):
Thanks!

PR #939, opened 2020-12-06
Fix intra rustdoc link todos

Folyd:

PR #940, opened 2020-12-07
Add a rustdoc job to CI

dtolnay:
@Folyd


dtolnay:
Confirmed that this fails against the first commit of #939. https://github.com/dtolnay/syn/runs/1508469062

PR #943, opened 2020-12-10
update `File` example program output

sharnoff:
The example listed in the documentation for `syn::File` had output that seems to be for a previous version of the code in the example - this just updates that.


dtolnay (approved):
Thanks!

Issue #857, opened 2020-06-22
Make pat::parsing::multi_pat_with_leading_vert public in some form


udoprog:
This is used in the [`Parse` impl for `Arm`](https://github.com/dtolnay/syn/blob/80c6889684da7e0a30ef5d0b03e9d73fa6160541/src/expr.rs#L2586), and would be useful when building custom parsing inside and around match statements.

CC: udoprog/genco#12

dtolnay:
This is now available as [`Pat::parse_multi_with_leading_vert`](https://docs.rs/syn/2/syn/enum.Pat.html#method.parse_multi_with_leading_vert).

PR #947, opened 2020-12-31
Make parse_stmt(input, true) available as pub

lahwran:
Hey yall - I'm writing a grammar where I needed to copy and modify Block::parse_within, which calls these two currently-private functions. I exposed them as methods, since that seems to be how the library is organized.

Needs further documentation before merge - can anyone contribute that? Also I'm not 100% sure about the api I'm exposing, so I'd love api review. This can move to an issue if that's appropriate.

Edit: also, if this is just not even the right thing to do, lmk. I figured I'd just write it since I needed it and then PR it in case y'all agreed, but nbd if it needs more discussion anyway.

(I'm likely to also pr an attempt to fix https://github.com/dtolnay/syn/issues/857 at some point, and I'd take inspiration from comments on this one.)

### DIFF
--- a/src/expr.rs
+++ b/src/expr.rs
@@ -1023,19 +1023,25 @@ ast_enum! {
 
 #[cfg(any(feature = "parsing", feature = "printing"))]
 #[cfg(feature = "full")]
-pub(crate) fn requires_terminator(expr: &Expr) -> bool {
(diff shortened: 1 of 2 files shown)


lahwran:
Closing due to inactivity.

PR #969, opened 2021-02-23
De-duplicate the generated code for `surround`

Marwes:
Reduces the output of "llvm lines" by ~1%

### DIFF
--- a/src/data.rs
+++ b/src/data.rs
@@ -433,18 +433,14 @@ mod printing {
     #[cfg_attr(doc_cfg, doc(cfg(feature = "printing")))]
     impl ToTokens for FieldsNamed {
         fn to_tokens(&self, tokens: &mut TokenStream) {
-            self.brace_token.surround(tokens, |tokens| {
-                self.named.to_tokens(tokens);
-            });
+            self.brace_token.surround_tokens(tokens, &self.named);
         }
     }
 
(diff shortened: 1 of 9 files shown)


botika:
@Marwes `mem:take(taked)` can be replace by `mem::replace(taked, Default::default())` in `rust 1.31`

Marwes:
Fixed

botika:
@Marwes  I'm nothing in this repo, I just wanted to help. it is the owner who has to approve it. I don't know if the owner like it or not. Sometimes, is quite busy. 

Sorry I did not mean to confuse.

botika (approved):

dtolnay (changes requested):
This is more complexity than I would be willing to add for that improvement. Is it possible to do this with ordinary function calls, instead of doing things through a guard object? Also I would prefer not to introduce any new `pub` API on the delimiter token types.

Marwes:
Fixed, though `pub(crate)` couldn't be used because of compatibility with old rustc.

dtolnay (commented):
I think ultimately I would prefer not to accept this. Adding new public API surface area solely for a 1% LLVM reduction is not a good tradeoff, and rewriting things in terms of a private API for a 1% LLVM reduction is also not good at the expense of those impls no longer being copyable and adaptable into user code.

Thanks anyway for the PR!

PR #970, opened 2021-02-23
Replace assertions in Puctuated with more helpful panic messages

osa1:
Debugging proc macros is very difficult as we can't easily get
backtraces into the proc macro code. Improved error messages will
hopefully give some hint to the user on what the problem is.


botika:
Can use https://doc.rust-lang.org/std/macro.assert.html like `assert!(some, format_args)`. I think it's more correct than `panic!`.

osa1:
Huh.. I didn't even know `assert!` can take multiple arguments. I updated the PR.

dtolnay (approved):
Thanks!

PR #1017, opened 2021-04-10
Ensure that README code example is up to date

GuillaumeGomez:
This provide a simple way to check that your README example is up-to-date.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -306,6 +306,13 @@ extern crate unicode_xid;
 #[cfg(feature = "printing")]
 extern crate quote;
 
+#[cfg(doctest)]
+#[macro_use]
+extern crate doc_comment;
+
+#[cfg(doctest)]
+doctest!("../README.md");
+
 #[macro_use]
 mod macros;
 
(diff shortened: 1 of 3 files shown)


src/lib.rs
```diff
@@ -306,6 +306,13 @@ extern crate unicode_xid;
 #[cfg(feature = "printing")]
 extern crate quote;
 
+#[cfg(doctest)]
```
taiki-e, line 309: Note that this increases MSRV to 1.40. https://github.com/BurntSushi/termcolor/issues/35

GuillaumeGomez: Outch, good point!

dtolnay (commented):
I would prefer not to make this change. You've sent equivalent PRs to some of my other projects in the past, which were similarly rejected. Could you please avoid sending any more of these to all projects under my account?

GuillaumeGomez:
Sorry, was using syn in one of my project and realized that, completely forgot I already sent the same change proposal to other project of yours. Will try to not do it again.

PR #1025, opened 2021-04-27
Parse type parameter defaults on impl blocks

taiki-e:
rustc syntactically accepts this.

```rust
#[cfg(any())]
impl<T = ()> () {}
```

[playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=5f585898bde387b258cecc57558e9aa9)

### DIFF
--- a/src/item.rs
+++ b/src/item.rs
@@ -2469,7 +2469,8 @@ pub mod parsing {
                 || (input.peek2(Ident) || input.peek2(Lifetime))
                     && (input.peek3(Token![:])
                         || input.peek3(Token![,])
-                        || input.peek3(Token![>]))
+                        || input.peek3(Token![>])
+                        || input.peek3(Token![=]))
                 || input.peek2(Token![const]));
         let mut generics: Generics = if has_generics {
             input.parse()?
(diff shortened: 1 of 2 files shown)


taiki-e:
Oh, actually, rustc accepts this semantically, but treats it as deprecated. I'm not sure if this is planned to be eventually rejected syntactically as well. However, I feel this is useful when writing proc-macro if this is syntactically accepted. (https://github.com/taiki-e/easy-ext/pull/32).

```text
error: defaults for type parameters are only allowed in `struct`, `enum`, `type`, or `trait` definitions.
 --> src/main.rs:2:6
  |
2 | impl<T = ()> () {}
  |      ^
  |
  = note: `#[deny(invalid_type_param_default)]` on by default
  = warning: this was previously accepted by the compiler but is being phased out; it will become a hard error in a future release!
  = note: for more information, see issue #36887 <https://github.com/rust-lang/rust/issues/36887>
```

dtolnay (approved):
Thanks! Yeah I am generally on board with parsing anything accepted syntactically by rustc that has any remote semblance of usefulness.

PR #1026, opened 2021-04-27
Use Option::get_or_insert_with in Generics::make_where_clause

taiki-e:
https://github.com/dtolnay/syn/blob/e6f9080447f3663890abcb469f52a35a9beb6a58/src/generics.rs#L174-L176


dtolnay (approved):
Thanks!

PR #1030, opened 2021-05-07
Add FromIterator<Field> for Fields

gardspirito:
Allows to create Fields from Field iterator.


gardspirito:
Bump?..

dtolnay (commented):
I would prefer not to build this into this crate. Thanks anyway for the PR!

PR #1033, opened 2021-05-21
example: rename `try_main` to `main`

Hofer-Julian:
This example was probably written before Rust 1.26 was introduced.
Since then,  main can return Result
https://github.com/rust-lang/rust/blob/master/RELEASES.md#version-1260-2018-05-10

### DIFF
--- a/examples/dump-syntax/src/main.rs
+++ b/examples/dump-syntax/src/main.rs
@@ -22,11 +22,11 @@ use std::env;
 use std::ffi::OsStr;
 use std::fmt::{self, Display};
 use std::fs;
-use std::io::{self, Write};
+use std::io;
 use std::path::{Path, PathBuf};
-use std::process;
 use colored::Colorize;
 
+#[derive(Debug)]
 enum Error {
     IncorrectUsage,
     ReadFile(io::Error),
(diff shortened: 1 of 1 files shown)


dtolnay (commented):
It looks like this does not compile.

```console
warning: unused import: `std::process`
  --> src/main.rs:27:5
   |
27 | use std::process;
   |     ^^^^^^^^^^^^
   |
   = note: `#[warn(unused_imports)]` on by default

warning: unused import: `Write`
  --> src/main.rs:25:21
   |
25 | use std::io::{self, Write};
   |                     ^^^^^

error[E0277]: `Error` doesn't implement `Debug`
  --> src/main.rs:56:14
   |
56 | fn main() -> Result<(), Error> {
   |              ^^^^^^^^^^^^^^^^^ `Error` cannot be formatted using `{:?}`
   |
   = help: the trait `Debug` is not implemented for `Error`
   = note: add `#[derive(Debug)]` or manually implement `Debug`
   = note: required because of the requirements on the impl of `Termination` for `Result<(), Error>`
```

Hofer-Julian:
Weird...
Will have a look later

Hofer-Julian:
With the current commit it should be fine :)
Seems like I forgot to wait for `cargo check` to finish.

Hofer-Julian:
Ah but the behaviour changes that way.
Will close the PR and try another time again when I have more time

PR #1036, opened 2021-06-01
Add `bail!` for easy erroring in parsers

Kestrer:
I find myself very frequently writing code of the form:

```rust
return Err(parse::Error::new_spanned(
    item,
    format_args!("An error message: {}", s),
));
```

I think it would be very nice to have a macro to write this shorter:

```rust
bail!(item, "An error message: {}", s);
```

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -293,6 +293,61 @@ pub fn new2<T: Display>(start: Span, end: Span, message: T) -> Error {
     }
 }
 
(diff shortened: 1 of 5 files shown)


dtolnay (commented):
Thanks for the PR. I would prefer not to build this into syn, but it would be reasonable to put a macro like this into whichever crate it would be frequently used in.

PR #1060, opened 2021-09-03
Fix precedence of attrs on assignment expr in stmt position

dtolnay:


twitchyliquid64:
Heya, we think this may have caused a breakage due to a differing AST for attributes preceding a assignment expression. Is this expected / WAI?

https://github.com/twitchyliquid64/usbd-hid/issues/32

PR #1067, opened 2021-09-22
Switch non exhaustive syntax tree enums to use #[non_exhaustive]

dtolnay:
This follows through on the comments on the `__TestExhaustive` variants promising that their purpose will be substituted with a `deny(reachable)` rustc lint once one is available.

https://github.com/dtolnay/syn/blob/ac68efab9576ab0f79633a48ebee5cca7321ddde/src/expr.rs#L247-L251

That lint is now landing as `non_exhaustive_omitted_patterns` in Rust 1.57. The correct way to use it for checking exhaustivity of a match is:

```rust
match expr {
    Expr::Array(e) => {...}
    Expr::Assign(e) => {...}
    ...
    Expr::Yield(e) => {...}

    #[cfg_attr(test, deny(non_exhaustive_omitted_patterns))]
    _ => { /* some sane fallback */ }
}
```

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -19,6 +19,10 @@ fn main() {
         println!("cargo:rustc-cfg=syn_no_const_vec_new");
     }
 
+    if compiler.minor < 40 {
+        println!("cargo:rustc-cfg=syn_no_non_exhaustive");
+    }
+
     if compiler.minor < 56 {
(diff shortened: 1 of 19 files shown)


dtolnay:
Closing, to be reopened when 1.57.0 stable is out.

PR #1091, opened 2021-11-10
Extend<Pair<&..>> and Pairs::cloned supports for Punctuated

dhardy:
This is designed to make it easy to extend a `Punctuated` from an iterator over another `Punctuated`:

1. Auto-insert extra punctuation if extend-target does not have trailing comma. This is a breaking change due to the extra `Default` bound, so I may need to remove it.
2. Support extension from reference types.
3. Support `cloned` over a `Pairs` iterator.

The last two are redundant with each other, from the POV of the motivation.

### DIFF
--- a/src/gen_helper.rs
+++ b/src/gen_helper.rs
@@ -21,7 +21,7 @@ pub mod fold {
         }
     }
 
-    impl<T, U> FoldHelper for Punctuated<T, U> {
+    impl<T, U: Default> FoldHelper for Punctuated<T, U> {
         type Item = T;
         fn lift<F>(self, mut f: F) -> Self
         where
--- a/src/punctuated.rs
+++ b/src/punctuated.rs
@@ -614,6 +614,57 @@ impl<'a, T, P> Clone for Pairs<'a, T, P> {
     }
 }
 
+// Iterator::cloned is not usable on this type
+impl<'a, T: Clone, P: Clone> Pairs<'a, T, P> {
+    pub fn cloned(self) -> ClonedPairs<Self> {
+        ClonedPairs(self)
+    }
+}
+
+/// An iterator that clones the contents of an underlying iterator
+///
+/// This struct is created by [`Pairs::cloned`].
+pub struct ClonedPairs<I>(I);
+
+impl<'a, I, T: 'a, P: 'a> Iterator for ClonedPairs<I>
+where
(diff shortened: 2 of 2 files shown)


dhardy:
BTW I'm waiting on comment before looking at CI failures. Quite possibly the `Default` bound needs to be removed?

src/punctuated.rs
```diff
@@ -455,20 +466,22 @@ where
     }
 }
 
-impl<T, P> FromIterator<Pair<T, P>> for Punctuated<T, P> {
+impl<T, P: Default> FromIterator<Pair<T, P>> for Punctuated<T, P> {
```
dtolnay, line 469: This bound should not be needed.

src/gen_helper.rs
```diff
@@ -21,7 +21,7 @@ pub mod fold {
         }
     }
 
-    impl<T, U> FoldHelper for Punctuated<T, U> {
+    impl<T, U: Default> FoldHelper for Punctuated<T, U> {
```
dtolnay, line 24: This bound should not be needed.

src/punctuated.rs
```diff
@@ -976,6 +1066,36 @@ where
...
 }
 
+#[cfg(feature = "clone-impls")]
+#[cfg_attr(doc_cfg, doc(cfg(feature = "clone-impls")))]
+impl<'a, T: 'a, P: 'a> Pair<&'a T, &'a P>
+where
+    T: Clone,
+    P: Clone,
+{
+    fn cloned(&self) -> Pair<T, P> {
```
dtolnay, line 1076: This is the wrong signature for a `cloned` function. It would be unusual for someone to be dealing with a `&Pair<&T, &P>`. They would either have a `&Pair<T, P>` or a `Pair<&T, &P>`.

Compare to https://doc.rust-lang.org/1.65.0/std/option/enum.Option.html#method.cloned or https://doc.rust-lang.org/1.65.0/std/result/enum.Result.html#method.cloned or https://doc.rust-lang.org/1.65.0/std/ops/enum.Bound.html#method.cloned.

dtolnay (changes requested):

dtolnay:
I picked up two of these changes in #1255 and #1256. Thanks!

The rest is too far into diminishing returns I think. For example `.map(Pair::cloned)` supersedes the need for a new `ClonedPairs` iterator type.

Issue #869, opened 2020-07-20
accept_as_ident does not reject Rust 2018 keywords


rylev:
When parsing identifiers, `accept_as_ident` is used to reject those strings which are keywords. This does not include Rust 2018 keywords such as `async` and `try`. Additionally, the linked resource for keywords has moved and can now be found [here](https://doc.rust-lang.org/reference/keywords.html).

PR #1098, opened 2021-11-24
Add async, await, try to list of reserved keywords for ident and update docs link

tomlinton:
Closes https://github.com/dtolnay/syn/issues/869


hlopko:
Could we revive this PR? Assuming we can fix the tests, would it be accepted?

hlopko:
`dyn` may have become a keyword in 2021, should we add it as well?

dtolnay (approved):
Thanks!

PR #1106, opened 2021-12-15
Tolerate the standard library reusing ThreadIds

dtolnay:
Allegedly, under some team members' interpretation of https://doc.rust-lang.org/1.57.0/std/thread/struct.ThreadId.html, the wording *"A ThreadId has a unique value for each thread that creates one"* is ambiguous as to whether two different threads in the same process might have the same "unique value" for their `ThreadId`, as long as the two threads do not exist at the same time.

Failure mode:

- Create a syn::Error with Span on thread 1
- Pass the error over to thread 2
- Thread 1 ends
- Create thread 3, which gets assigned the same `ThreadId` as thread 1
- Pass the error over to thread 3
- Read its Span, which accesses thread_local span table from dead thread 1

https://github.com/rust-lang/rust/pull/84083

### DIFF
--- a/src/thread.rs
+++ b/src/thread.rs
@@ -1,5 +1,6 @@
 use std::fmt::{self, Debug};
-use std::thread::{self, ThreadId};
+use std::num::NonZeroUsize;
+use std::sync::atomic::{AtomicUsize, Ordering};
 
 /// ThreadBound is a Sync-maker and Send-maker that allows accessing a value
 /// of type T only from the original thread on which the ThreadBound was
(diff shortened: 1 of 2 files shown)


src/thread.rs
```diff
@@ -1,5 +1,11 @@
 use std::fmt::{self, Debug};
-use std::thread::{self, ThreadId};
+use std::sync::atomic::Ordering;
+
+#[cfg(not(syn_no_atomic_u64))]
+use std::{num::NonZeroU64 as NonZeroThreadId, sync::atomic::AtomicU64 as AtomicThreadId};
```
cuviper, line 5: Beware, some 32-bit targets do not have `Atomicu64`!
file:///usr/share/doc/rust/html/std/sync/atomic/index.html#portability

dtolnay: I don't anticipate having to land this. :wink:

dtolnay (commented):
Closing in light of the consensus in https://github.com/rust-lang/rust/pull/84083. ThreadIds are guaranteed not to be reused even after a thread finishes.

PR #1131, opened 2022-02-17
Added a note on generics struct

JohnScience:
The links to the reference do not fully compensate for the lack of examples yet they allow even less experienced rust developers understand which items are represented by [Generics](https://docs.rs/syn/latest/syn/struct.Generics.html) struct.


dtolnay (approved):
Thanks!

PR #1133, opened 2022-02-28
Collect an Iterator<Item = syn::Result<T>> into syn::Result<FromIterator<T>> whilst combining collected syn::Errors

GothAck:
Been using this trait a lot recently, would be great if it were part of syn itself :)


dtolnay (commented):
Thanks for the PR! I can see how this would be useful but I would prefer not to build this into this library.

Issue #1143, opened 2022-03-14
Fails to parse a `use` statement with leading colons inside a `UseTree`


danielhenrymantilla:
I tried this code:

```rust
use {
    ::{
        proc_macro2::{
            TokenStream,
        },
        syn::*,
    },
    crate::{
        utils::{
            // …
        },
    }, 
};

fn main() {
    parse_file(include_str!(concat!(env!("CARGO_MANIFEST_DIR"), "/", file!())))
        .unwrap()
    ;
}
```

I expected to see this happen: `parse_file` succeeding.

Instead, it fails because of the leading colons inside the braced use "trees".

  - [Playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=3d33ca344bdbed255155ee9b6fde7b2b)

___

While fixing it ought to be easy enough for me to submit a PR, I'm struggling to come up with a fix that does not involve:

  - a lossy process (parsing the leading colon and then discarding it by lack of field to assign it to);
  - a breaking change (changing the structure of `ItemUse` or of its `UseTree`).

The only way I personally see would be to define a new `ItemUse2` variant of `Item`, and somehow have the parser fall back to `ItemUse2` while it is parsing a `ItemUse` if it finds out there is a leading colon inside the first braced `UseTree`…

___

I've discovered this bug by incidentally coming up with an example that https://github.com/dtolnay/prettyplease can't handle because of it.

dtolnay:
I think it's fine not to parse this. No one should write code like that, so making the syntax tree more complicated for it would be a bad tradeoff.

danielhenrymantilla:
A pity, I have to admit, but yeah, I understand that the _tradeoff_ may not warrant the ordeal that _full parse/AST support_ would suppose while under the `1.0` semver constraints[^1].

[^1]: I think it could be a nice thing to change for `2.0`, should such a release ever happen

___

### Do not parse, just _tolerate_ (example of a possible implementation in https://github.com/dtolnay/syn/pull/1145)

Then, what about dropping the full parsing/AST support, but trying to keep **a minimum amount of support** so that `prettyplease` or other stuff parsing whole files/modules can **tolerate/withstand such a `use`**.

Indeed, I don't think that occurrences such code would be _that_ rare, or even rare enough for `syn` to be able to afford this inconsistency with `rustc` parsing: some basic grep.app searches do show occurrences of both:

[...]

PR #1145, opened 2022-03-15
Tolerate leading colons in paths inside a braced use item

danielhenrymantilla:
but this does not really *parse* them in the AST, since it would be a subtle
and complex thing to do without breaking the 1.0 semver, thus not worth the trouble.

Tackles https://github.com/dtolnay/syn/issues/1143 (see that issue for more info/context about this PR).

### DIFF
--- a/src/item.rs
+++ b/src/item.rs
@@ -1024,7 +1024,23 @@ pub mod parsing {
                     Err(lookahead.error())
                 }
             } else if lookahead.peek(Token![use]) {
-                input.parse().map(Item::Use)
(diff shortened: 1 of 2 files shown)


dtolnay (commented):
It looks like this is just silently ignoring *any* syntax errors inside the braced part of a `use`, which I don't think would be an acceptable behavior.

Overall I would still prefer to make no change here. From your comments in the linked issue it sounds like you were primarily interested in this because of `prettyplease`, but I think this PR would not help you with `prettyplease` because that crate does not work on anything containing Item::Verbatim.

Issue #1149, opened 2022-03-21
Add more example of how to use "fold"


titaneric:
I would like to share more example of fold in the docs which is  to "publicize" the main function. The similar technique could be used to import the "publicized function" from `lib.rs` into the `main.rs` of the library crates which is possible to accelerate the compiling speed by the power of `sccache`.

Example:

Suppose we have a library crates called `pub_fn`

Cargo.toml
```toml
[package]
name = "pub_fn"
version = "0.1.0"
edition = "2021"
```

original lib.rs
```rs
fn main() {
   println!("print from main fn");
}
```

transformed lib.rs
```rs
pub fn main() {
   println!("print from main fn");
}
```

main.rs
```rs
fn main() {
  pub_fn::main();
}
```

dtolnay:
Thanks for the suggestion! I'll close this per https://github.com/dtolnay/syn/pull/1150#pullrequestreview-921231740.

PR #1150, opened 2022-03-21
Add pub main fn example for fold

titaneric:
As #1149 , I would like to add more example of fold function in syn.

This example demonstrate how to use fold to transform the given function into public visibility.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -775,6 +775,68 @@ mod gen {
     ///     // Output: (((a)()) + (((b)((1))) * ((c).d)))
     /// }
     /// ```
+    ///
+    /// This fold publicize the given main function.
+    ///
+    /// ```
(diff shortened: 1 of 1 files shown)


dtolnay (commented):
Thank you for the PR!

I don't think the use case described here would be a good example for `Fold` because it's far easier to accomplish this without using `Fold`.

```rust
let mut file: syn::File = syn::parse_str(SRC_CODE).unwrap();
for item in &mut file.items {
    if let Item::Fn(item) = item {
        if item.sig.ident == "main" {
            item.vis = parse_quote!(pub);
        }
    }
}
println!("{}", quote!(#file));
```

titaneric:
Oh, I did not expect how easy it could, thank you for sample code.

Issue #1157, opened 2022-04-05
`RangeLimits` does not implement `ToTokens`


frank2:
Hello! This is a very simple issue I noticed. The `RangeLimits` enum doesn't implement `ToTokens`, which is out of step with the rest of the enums in the library. I'm assuming this is an oversight and not intentional since so many other enums implement `ToTokens`.

Since this is such a simple fix I'll follow up with a PR to fix the issue!

frank2:
This is merged, so I'm gonna close it!

PR #1158, opened 2022-04-05
#1157 implement ToTokens for RangeLimits enum

frank2:
Fixes a small issue where the `RangeLimits` enum didn't implement `ToTokens` when it probably should have.


dtolnay (approved):
Thanks!

PR #1162, opened 2022-04-23
Added `dyn` to Type::TraitObject docs

SOF3:
The old syntax is now a compile error, and it makes it difficult for users to search when they are looking for the keyword "dyn" on the documentation.


dtolnay (approved):
Thanks!

PR #1164, opened 2022-04-25
add yeet syntax support

conradludgate:
Implementation of https://github.com/rust-lang/rust/issues/96373

Not sure on how syn approaches experimental syntax, so feel free to close if this is out of scope for now

### DIFF
--- a/src/expr.rs
+++ b/src/expr.rs
@@ -222,6 +222,9 @@ ast_enum_of_structs! {
         /// A while loop: `while expr { ... }`.
         While(ExprWhile),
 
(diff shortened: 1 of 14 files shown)


dtolnay (commented):
I am not interested in adding a permanent syntax tree node for this.

Thanks anyway for the PR.

scottmcm:
Given that this is still explicitly experimental, not adding it to `syn` makes sense to me 👍

(It might make sense once there's an accepted RFC with a real keyword, even if it's not yet stable at that point, but when it's just an exploration I wouldn't want to force this on the library forever.)

PR #1171, opened 2022-05-09
Use Some from $crate

zakarumych:
Fixes build when `Some` symbol is redefined in scope where `custom_keyword` macro is invoked.


dtolnay (approved):
Thanks!

Issue #1159, opened 2022-04-07
Is there any plan to fully support let-else?


TennyZhuang:
Currently, let_else will be treated as `ExprVerbatim`:

```
[tests/test_stmt.rs:108] &stmt = Ok(
    Semi(
        Verbatim(
            TokenStream [
                Ident {
                    sym: let,
                },
                Ident {
                    sym: Some,
                },
                Group {
                    delimiter: Parenthesis,
                    stream: TokenStream [
                        Ident {
                            sym: x,
                        },
                    ],
                },
                Punct {
                    char: '=',
                    spacing: Alone,
                },
                Ident {
                    sym: None,
                },
                Ident {
                    sym: else,
                },
                Group {
                    delimiter: Brace,
                    stream: TokenStream [
                        Ident {
                            sym: yield,
                        },
                        Literal {
                            lit: 1,
                        },
                        Punct {
                            char: ';',
                            spacing: Alone,
                        },
                    ],
                },
            ],
        ),
        Semi,
    ),
)
```

I'd like to contribute to it! But I'm not sure whether it will be accepted for an unstable feature

TennyZhuang:
It seems that we can add a new field `else_branch: Box<Expr>` in `struct Local`

wkordalski:
It looks reasonable.
I think it is the right moment to do so, as let-else get into stable in next month (i.e. it is in beta now).

PR #1231, opened 2022-10-11
Support full let-else statement parsing

wkordalski:
Currently let-else statements are parsed as `ExprVerbatim` (i.e. they are not parsed internally).

*let-else* statements are in beta now, so they will most probably appear in stable in November. Therefore, I think, now is the right moment for adding full parsing implementation.

This PR adds such parsing implementation to syn.
Fixes: https://github.com/dtolnay/syn/issues/1159

### DIFF
--- a/src/stmt.rs
+++ b/src/stmt.rs
@@ -30,7 +30,7 @@ ast_enum! {
 }
 
 ast_struct! {
-    /// A local `let` binding: `let x: u64 = s.parse()?`.
+    /// A local `let` binding: `let x: u64 = s.parse()?` (including let-else construction).
     ///
     /// *This type is available only if Syn is built with the `"full"` feature.*
     #[cfg_attr(doc_cfg, doc(cfg(feature = "full")))]
@@ -38,8 +38,23 @@ ast_struct! {
         pub attrs: Vec<Attribute>,
         pub let_token: Token![let],
         pub pat: Pat,
-        pub init: Option<(Token![=], Box<Expr>)>,
+        pub init: Option<LocalInitializer>,
         pub semi_token: Token![;],
+    }
+}
+
+ast_struct! {
+    /// An initializer of local `let` binding (with optional diverging else-block)
+    ///
+    /// `LocalInitializer` represents `= s.parse()?` in `let x: u64 = s.parse()?`
+    /// and `= r else { return }` in `let Ok(x) = r else { return }`.
+    ///
+    /// *This type is available only if Syn is built with the `"full"` feature.*
+    #[cfg_attr(doc_cfg, doc(cfg(feature = "full")))]
+    pub struct LocalInitializer {
+        pub eq_token: Token![=],
+        pub expr: Box<Expr>,
+        pub else_block: Option<(Token![else], Box<ExprBlock>)>,
     }
 }
 
@@ -47,7 +62,7 @@ ast_struct! {
 pub mod parsing {
     use super::*;
     use crate::parse::discouraged::Speculative;
-    use crate::parse::{Parse, ParseBuffer, ParseStream, Result};
+    use crate::parse::{Parse, ParseStream, Result};
     use proc_macro2::TokenStream;
 
     struct AllowNoSemi(bool);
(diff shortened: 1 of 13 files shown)


deg4uss3r:
Hey everyone! Now that let/else is stable I'd love to try to get this PR in, @wkordalski is there anything I can help you with to get CI to pass? 

@dtolnay does this PR pass your initial review? Is there another direction you'd like to take (or another PR that I missed in the queue)?

wkordalski:
Hi!
I miss let-else support very much and wait for adding its support to `syn`.
I need some information on the workflow of adding such features to `syn`, so that tests will pass on CI.
And I hope that someone (@dtolnay?) can give me some hints.

## The CI problem

One of the tests is to run `syn`'s internal codegen.
The codegen at the end of its work formats code using `prettyplease`, which internally uses `syn` to parse code to be formatted.
In the failing syn test (that runs syn's internal codegen) we compile `prettyplease` crate [with overwritten `syn` version](https://github.com/wkordalski/syn/blob/ddcaf7341c2e813949dcce0140f451e1860aecc5/codegen/Cargo.toml#L27) (i.e. `syn` from this branch (with let-else support), not the recently *released* version). However current version of `prettyplease` is not compatible with "`syn` from this branch" and it fails to compile.

We could update `prettyplease` to be compatible with "`syn` from this branch", but then `prettyplease` won't be compatible with currently-released version of `syn`, and `prettyplease` CI will fail.

## Simplest solution

I think the simplest solution is:
1. Add to `prettyplease` information that it supports `syn` up to version "currently-released-version" (via Cargo.toml)
2. Remove the overwrite of `syn` crate in codegen (simply remove [this line](https://github.com/wkordalski/syn/blob/ddcaf7341c2e813949dcce0140f451e1860aecc5/codegen/Cargo.toml#L27)). Then codegen from "`syn` from this branch" will happily use `prettyplease` that will use "currently-released-version" of `syn`.
3. Now all tests from `syn` CI pass, and we can merge this PR and publish (possibly as prerelease version).
4. Update `prettyplease` to use and compile happily with the prerelease version of `syn` (also modify the version requiremest in Cargo.toml of `prettyplease`)
5. The CI of `prettyplease` should pass, so we can publish it.
6. Finally, we can publish the prerelease version of `syn` as normal release.

### Some other thoughts

- We can skip the publishing prerelease version step by stating in `prettyplease`'s Cargo.toml to use `syn` from this branch, but I feel this won't be the better option.
- Should we make minor/major version bump as we change publicly-accessible data structures, what is a breaking change?
- The "simplest solution" is quite complicated and I would love to hear from maintainers of this crate if there is another solution that is simpler.

wkordalski:
Ok... Thanks to https://github.com/dtolnay/syn/commit/ec9de4642d70105d2a73f1eb6654405ce5ff8be5 all tests pass and we can move forward.
I would appreciate some review, esp. of the changes to data structures.
After merging this I can also update `prettyplease`.

wkordalski:
@dtolnay Can I have your review on this?

src/stmt.rs
```diff
@@ -246,17 +260,19 @@ pub mod parsing {
...
-                let content;
-                braced!(content in input);
-                content.call(Block::parse_within)?;
-                let verbatim = Expr::Verbatim(verbatim::between(begin, input));
-                let semi_token: Token![;] = input.parse()?;
-                return Ok(Stmt::Expr(verbatim, Some(semi_token)));
-            }
+            let else_block = if input.peek(Token![else]) {
+                let else_token = input.parse::<Token![else]>()?;
+                let expr_block = input.parse::<ExprBlock>()?;
```
dtolnay, line 265: Parsing ExprBlock here is incorrect because it'd incorrectly accept attributes and labels on the else-block, which is not valid syntax in Rust. Things like: `let true = false else 'label: { return; }; let true = false else #[foo] { return; };`

src/stmt.rs
```diff
@@ -338,11 +354,21 @@ mod printing {
...
-            if let Some((eq_token, init)) = &self.init {
-                eq_token.to_tokens(tokens);
+            if let Some(init) = &self.init {
                 init.to_tokens(tokens);
             }
             self.semi_token.to_tokens(tokens);
         }
     }
+
+    impl ToTokens for LocalInitializer {
```
dtolnay, line 364: This new impl is missing `#[cfg_attr(doc_cfg, doc(cfg(feature = "printing")))]`

dtolnay (changes requested):
Thanks!

dtolnay:
I am going to take care of the requested changes myself.

PR #1237, opened 2022-10-20
Restrict PartialOrd comparison to only tokens in the same buffer

dtolnay:
How does this look @CAD97? I was hesitant about directly exposing a pointer comparison across unrelated allocations.

```rust
use syn::buffer::TokenBuffer;
let a = TokenBuffer::new2("...".parse().unwrap());
let b = TokenBuffer::new2("...".parse().unwrap());
println!("{:?}", a.begin().partial_cmp(&b.begin()));
```

### DIFF
--- a/src/buffer.rs
+++ b/src/buffer.rs
@@ -26,7 +26,8 @@ enum Entry {
     Ident(Ident),
     Punct(Punct),
     Literal(Literal),
-    End,
+    // End entries contain the offset (negative) to the start of the buffer.
+    End(isize),
 }
 
 /// A buffer that can be efficiently traversed multiple times, unlike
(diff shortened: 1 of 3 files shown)


CAD97 (commented):
That looks really good to me! It's definitely nice to have the cursors for distinct buffers unordered rather than arbitrary-but-consistent order.

It's nice it also gives a purpose to the `End` entries rather than them solely existing to make implementation easier.

(And as a tertiary nicety, it doesn't necessarily preclude a rewrite of `Cursor` that uses slices and less `unsafe`, as such could still do the offset for comparison, just not access outside the cursor range.)

TLDR: r+

(Apparently, "Can not approve pull request without explicit repository access" — did GH decide to prevent drive-by PR approvals? Or is that an old setting I don't know about? Irrelevant, I guess; though it is nice to be able to give a structured ±1 without merge capabilities, it's definitely abusable.)

PR #1242, opened 2022-11-10
Implement Pair::map

dhardy:
Just a little feature I wanted.

### DIFF
--- a/src/punctuated.rs
+++ b/src/punctuated.rs
@@ -987,6 +987,17 @@ impl<T, P> Pair<T, P> {
             Pair::End(t) => (t, None),
         }
     }
+
+    /// Maps a `Pair<T, P>` to a `Pair<U, P>`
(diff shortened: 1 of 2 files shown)


dtolnay (commented):
Could you give a realistic example of code where this would be used?

dhardy:
https://github.com/kas-gui/impl-tools/blob/fe651f1970ab0ad70316607e8a73b7b7b07567a9/lib/src/generics.rs#L497

dtolnay (commented):
I think I'm not convinced that this would be broadly useful enough to be worth adding. In all the macros I've written, I don't recall it coming up hardly ever.

The code you linked is a fork of syn's Generics parser and printer, which is far from a typical use case. The `map` being done is a reimplementation of syn's "clone-impls" feature, which is also not a thing that downstream code would typically need to do. And even there, Pair::map as written in this PR would not apply because the `P` type is not being kept constant.

Issue #1246, opened 2022-12-01
Punctuated iterators are overly restrictive of lifetimes in Drop


dtolnay:
```rust
use syn::punctuated::Punctuated;

fn repro() -> Punctuated<i32, ()> {
    let punctuated = Punctuated::new();
    for element in &punctuated {
        if *element == 2 {
            return punctuated;
        }
    }
    panic!();
}
```

```console
error[E0505]: cannot move out of `punctuated` because it is borrowed
 --> src/main.rs:7:20
  |
5 |     for element in &punctuated {
  |                    -----------
  |                    |
  |                    borrow of `punctuated` occurs here
  |                    a temporary with access to the borrow is created here ...
6 |         if *element == 2 {
7 |             return punctuated;
  |                    ^^^^^^^^^^ move out of `punctuated` occurs here
8 |         }
9 |     }
  |     - ... and the borrow might be used here, when that temporary is dropped and runs the destructor for type `syn::punctuated::Iter<'_, i32>`
```

This is a false positive; the code above should be fine exactly as written. Equivalent code written against Vec, HashMap, BTreeMap all compile without error.

It has something to do with the use of `#[may_dangle]` in Vec's Drop impl &mdash; https://github.com/rust-lang/rust/blob/1.65.0/library/alloc/src/vec/mod.rs#L2915

PR #1248, opened 2022-12-01
Simplify dyn/impl parsing enabled by Punctuated improvements

dtolnay:
Cleanup of the code added in #1245. Enabled by #1247. This code is where I noticed #1246.

### DIFF
--- a/src/ty.rs
+++ b/src/ty.rs
@@ -925,24 +925,20 @@ pub mod parsing {
         ) -> Result<Punctuated<TypeParamBound, Token![+]>> {
             let bounds = TypeParamBound::parse_multiple(input, allow_plus)?;
             let mut last_lifetime_span = None;
-            let mut at_least_one_trait = false;
             for bound in &bounds {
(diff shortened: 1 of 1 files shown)


dtolnay (commented):
Apparently this only works on Rust 1.40+, which we aren't requiring yet.

dtolnay:
Older rustc:

```console
error[E0505]: cannot move out of `bounds` because it is borrowed
   --> src/ty.rs:930:59
    |
928 |   for bound in &bounds {
    |                -------
    |                |
    |                borrow of `bounds` occurs here
    |                a temporary with access to the borrow is created here ...
929 |       match bound {
930 |           TypeParamBound::Trait(_) => return Ok(bounds),
    |                                                 ^^^^^^ move out of `bounds` occurs here
...
935 |   }
    |   - ... and the borrow might be used here, when that temporary is dropped and runs the destructor for type `punctuated::Iter<'_, generics::TypeParamBound>`
```

PR #1267, opened 2022-12-13
Delete AttributeArgs type

dtolnay:
I am working on a replacement for this.

### DIFF
--- a/src/attr.rs
+++ b/src/attr.rs
@@ -425,44 +425,6 @@ ast_enum_of_structs! {
     }
 }
 
-/// Conventional argument type associated with an invocation of an attribute
-/// macro.
-///
-/// For example if we are developing an attribute macro that is intended to be
-/// invoked on function items as follows:
-///
-/// ```
-/// # const IGNORE: &str = stringify! {
-/// #[my_attribute(path = "/v1/refresh")]
-/// # };
-/// pub fn refresh() {
-///     /* ... */
-/// }
-/// ```
-///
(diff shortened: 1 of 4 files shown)


erichulburd:
What motivated deletion of `AttributeArgs`? Is there a design pitfall? Or was this crate not the right place for it?

I've been following https://developerlife.com/2022/03/30/rust-proc-macro/ to develop a proc macro. In the absence of `sync::AttributeArgs`, is there anything to be weary of if I declare a private type `Vec<NestedMeta>` and use it to parse my macro's attributes?

erichulburd:
Ahh - there is: https://docs.rs/syn/latest/syn/meta/struct.ParseNestedMeta.html

So something like this should work:


```rs
#[derive(Default)]
struct MyMacroConfiguration {
  my_key: String,
}

#[proc_macro_attribute]
pub fn my_macro(attr: TokenStream, item: TokenStream) -> TokenStream {
    // https://docs.rs/syn/latest/syn/struct.Attribute.html#method.parse_nested_meta
    let mut config = MyMacroConfiguration::default();
    let config_parser = syn::meta::parser(|meta| {
      if meta.path.is_ident("my_key") {
        let value: syn::LitStr = meta.value()?.parse()?;
        config.my_key = value.value();
        return Ok(())
     }
     let source = meta.path.get_ident().unwrap().span().source_text();
      Err(meta.error(format!("unrecognized meta {source}"))
    });
    parse_macro_input!(attr with config_parser);
    my_macro_impl(item, config)
}
```

PR #1281, opened 2022-12-22
Add `Item::attrs` to get the attributes of an item if available

### LINKED ISSUE #866: Add attrs() and attrs_mut() getters to syn::Item
While developing certain proc macros using `syn` I found getters that return the attribute slice for a generic `syn::Item` to be missing. The implementation is straight forward and simple, however, daunting:

## Example through some extension trait

```rust
impl Attrs for syn::Item {
    fn attrs(&self) -> &[syn::Attribute] {
        use syn::Item;
        match self {
            Item::Const(syn::ItemConst { attrs, .. })
            | Item::Enum(syn::ItemEnum { attrs, .. })
            | Item::ExternCrate(syn::ItemExternCrate { attrs, .. })
            | Item::Fn(syn::ItemFn { attrs, .. })
            | Item::ForeignMod(syn::ItemForeignMod { attrs, .. })
            | Item::Impl(syn::ItemImpl { attrs, .. })
            | Item::Macro(syn::ItemMacro { attrs, .. })
            | Item::Macro2(syn::ItemMacro2 { attrs, .. })
            | Item::Mod(syn::ItemMod { attrs, .. })
            | Item::Static(syn::ItemStatic { attrs, .. })
            | Item::Struct(syn::ItemStruct { attrs, .. })
            | Item::Trait(syn::ItemTrait { attrs, .. })
            | Item::TraitAlias(syn::ItemTraitAlias { attrs, .. })
            | Item::Type(syn::ItemType { attrs, .. })
            | Item::Union(syn::ItemUnion { attrs, .. })
            | Item::Use(syn::ItemUse { attrs, .. }) => attrs,
            _ => &[],
        }
    }
}
```

Note that a getter for this would also have the benefit of such dependent crates to not care if `syn` ever adds new variants. If `syn` provided this dependencies could simply make use of `syn` handling the update.

kraktus:
Hello, thanks for the crate!

When looking at `Item` I was surprised not to find APIs that ease its use, without having to pattern-match every time, like for [`rustc_hir::ItemKind`](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_hir/hir/enum.ItemKind.html). This PR add a convenience method to have an easy access to the attributes of an item. 

I don't know the policy about new APIs, so feel free to close if deemed out-of-scope. About the feature-flag to use and the documentation I can fix what's needed, when for `parsing` as a first guess.

As a side note `test_expr_size` and `test_type_size` fail on arm64 (M1).

### DIFF
--- a/src/item.rs
+++ b/src/item.rs
@@ -347,6 +347,32 @@ impl Item {
             _ => unreachable!(),
         }
     }
+
+    #[cfg(feature = "parsing")]
+    pub fn attrs(&self) -> Option<&Vec<Attribute>> {
+        match self {
+            Item::ExternCrate(ItemExternCrate { attrs, .. })
+            | Item::Use(ItemUse { attrs, .. })
+            | Item::Static(ItemStatic { attrs, .. })
+            | Item::Const(ItemConst { attrs, .. })
+            | Item::Fn(ItemFn { attrs, .. })
+            | Item::Mod(ItemMod { attrs, .. })
+            | Item::ForeignMod(ItemForeignMod { attrs, .. })
+            | Item::Type(ItemType { attrs, .. })
+            | Item::Struct(ItemStruct { attrs, .. })
+            | Item::Enum(ItemEnum { attrs, .. })
+            | Item::Union(ItemUnion { attrs, .. })
+            | Item::Trait(ItemTrait { attrs, .. })
+            | Item::TraitAlias(ItemTraitAlias { attrs, .. })
+            | Item::Impl(ItemImpl { attrs, .. })
+            | Item::Macro(ItemMacro { attrs, .. })
+            | Item::Macro2(ItemMacro2 { attrs, .. }) => Some(attrs),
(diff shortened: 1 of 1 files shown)


src/item.rs
```diff
@@ -347,6 +347,32 @@ impl Item {
             _ => unreachable!(),
         }
     }
+
+    #[cfg(feature = "parsing")]
+    pub fn attrs(&self) -> Option<&Vec<Attribute>> {
```
jf2048, line 352: I have a similar implementation of this in a crate here: [parse_attributes.rs#L336](https://github.com/jf2048/deluxe/blob/cdc455fe3090aaa7215fe7dd676c07814e195045/core/parse_attributes.rs#L336)

I was able to have it simply return `&[Attribute]`, and for `Item::Verbatim` it just returns `&[]`. Not sure if that is better or not.

mash180sx:
Why isn't this PR being merged?

kraktus:
gentle bump, 4 years later

PR #1349, opened 2023-01-31
Preserve c"..." and cr"..." literal tokens

dtolnay:
https://github.com/rust-lang/rfcs/pull/3348
https://github.com/rust-lang/rust/issues/105723


dtolnay:
Needs an implementation in proc-macro2's fallback parser as well.

PR #1351, opened 2023-02-03
Add note to `File` rustdoc about `parse_file`.

Lucretiel:
Mostly I'm doing this to make `parse_file` easy to discover with rustdoc hyperlinks


dtolnay (approved):
Thanks!

PR #1381, opened 2023-02-20
stricter parsing of string_continue escapes

ModProg:
According to https://doc.rust-lang.org/reference/tokens.html#string-literals
only these 4 characters are actually skipped.


dtolnay (approved):
Thanks!

PR #1385, opened 2023-02-26
doc: clarify that push_value can be used on empty Punctuated

SOF3:
I was briefly confused when I saw that it must push behind a trailing punctuation.


dtolnay (approved):
Thanks!

PR #1387, opened 2023-03-10
Fix from upstream Rust: map_in_place -> flat_map_in_place

philip-peterson:
This is just an upstream build breakage I saw.


dtolnay (commented):
This was fixed an hour and a half ago.

PR #1395, opened 2023-03-13
Add `Punctuated::remove(index)`

### LINKED ISSUE #1314: Method to remove values from Punctuated
The only current APIs on syn::punctuated::Punctuated that result in the Punctuated becoming smaller are `pop` (remove last) and `clear` (remove all).

https://github.com/dtolnay/syn/issues/343#issuecomment-657923539 mentioned needing to remove by index.

tguichaoua:
related to #1314

Add a `remove(index)` method that remove and return an element from `Punctated`.
This method is similar to [`Vec::remove(index)`](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.remove).

### DIFF
--- a/src/punctuated.rs
+++ b/src/punctuated.rs
@@ -238,6 +238,23 @@ impl<T, P> Punctuated<T, P> {
         }
     }
 
+    /// Removes and returns the punctuated pair at position `index`.
+    ///
+    /// # Panics
+    ///
+    /// Panics if `index` is out of bounds.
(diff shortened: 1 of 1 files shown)


tguichaoua:
CI fails because `syn` uses `quote` `1.0.25`, [which is yanked](https://crates.io/crates/quote/1.0.25)

notdanilo:
Wow. More than 6 months for a small PR.

PR #1400, opened 2023-03-15
Make eligible functions const.

tormeh:


dtolnay (commented):
I would prefer not to do this indiscriminately. If there is a specific use case that requires one of these to be usable in a const context, I am open to hearing about it.

PR #1407, opened 2023-03-17
Optimize syntax tree size of Punctuated (without allocation for empty Punctuated)

dtolnay:
Alternative attempt at #1406. Sadly this one has the same 20-30% improvement on syntax tree size and only 2-3% improvement on performance.


dtolnay:
I guess it's good news that performance appears not to be sensitive to syntax tree size, even for quite large nodes (200–300 bytes).

PR #1410, opened 2023-03-19
dd Punctuated::remove(index)

tguichaoua:
related to #1314

Add a `drain(range)` method to `Punctuated` that is similar to [`Vec::drain(range)`](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.drain).

### DIFF
--- a/src/punctuated.rs
+++ b/src/punctuated.rs
@@ -26,7 +26,8 @@ use std::fmt::{self, Debug};
 use std::hash::{Hash, Hasher};
 #[cfg(any(feature = "full", feature = "derive"))]
 use std::iter;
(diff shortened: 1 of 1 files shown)


dtolnay (commented):
Could you share the use case you have for this function?

tguichaoua:
> Could you share the use case you have for this function?

I don't have a specific use case in mind. I added it to more convenient than calling `remove` (https://github.com/dtolnay/syn/pull/1395) multiple times to remove a range.

dtolnay (commented):
Okay, I am not interested in adding a function without a use case.

Issue #1427, opened 2023-03-27
cannot find macro `compile_error` in this scope


smoelius:
A custom inner attribute (specifically, [this one](https://github.com/trailofbits/cast_checks)) applied at the crate root produces an error message like this:
```
error: cannot find macro `compile_error` in this scope
 --> src/main.rs:2:1
  |
2 | / #![feature(custom_inner_attributes, proc_macro_hygiene)]
3 | |
4 | | #![cast_checks::enable]
5 | |
6 | |
7 | | fn main() {
8 | | }
  | |_^
  |
  = help: consider importing one of these items:
          std::compile_error
          core::compile_error
```
I think the reason for the error message is that `compile_error!` is not qualified here:
https://github.com/dtolnay/syn/blob/e19a0fbfe1e035ec0b8e5915179b0035e398a257/src/error.rs#L283
In other words, I think it should be `std::compile_error!`.

To be clear, I don't think that making this error go away will make the crate-level, custom inner attribute work---I think there are other problems.

But making this error go away would help to expose those other problems and make them easier to diagnose.

Would you agree?

I could submit a PR, but I would need a pointer or two on how to change the referenced code. My attempts to add `std::` thus far have been unsuccessful.

PR #1431, opened 2023-03-30
Qualify `compile_error!`

smoelius:
Proposed fix for #1427

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -278,8 +278,19 @@ impl ErrorMessage {
             None => (Span::call_site(), Span::call_site()),
         };
 
-        // compile_error!($message)
+        // std::compile_error!($message)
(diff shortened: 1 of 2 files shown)


src/error.rs
```diff
@@ -278,8 +278,19 @@ impl ErrorMessage {
             None => (Span::call_site(), Span::call_site()),
         };
 
-        // compile_error!($message)
+        // std::compile_error!($message)
         TokenStream::from_iter(vec![
+            TokenTree::Ident(Ident::new("std", start)),
```
taiki-e, line 283: Please use `core` instead of `std`. Otherwise, it will generate an ugly error message in `#![no_std]` crate.

dtolnay (approved):
Thanks!

PR #1444, opened 2023-04-13
fixed the case when `TypeTuple::to_tokens` will result in `TypeParen`

Fancyflame:
When `TypeTuple` has only one argument but has no trailing comma, it will be considered as `TypeParen`. I fixed the code as `ExprTuple` did (Maybe you forgot to modify `TypeTuple`).


dtolnay (approved):
Thanks!

PR #1496, opened 2023-09-02
Add `require_ident` to `Path`

Fancyflame:
Parsing options inside a list is very commonplace:
```text
#[my_attr(path(linux = "path/to/file")]
          ^^^^ ^^^^^
          Here is our options
```

Although we have `get_ident`, we (at least I) would not like to handle `None`. Just like `require_path`, `require_list` and `require_key_value`, which I love.

```rust
match &*path.require_ident()?.to_string() {
    "linux" => { ... }
    "windows" => { ... }
    "macos" => { ... }
    other => return Err(Error::new(
        Span::call_site(),
        format!("platform `{other}` not supported")
    ))
}
```

### DIFF
--- a/src/path.rs
+++ b/src/path.rs
@@ -82,6 +82,26 @@ impl Path {
             None
         }
     }
+
+    /// Error if this does not contain only one segment.
+    ///
+    /// A path is considered an ident if:
+    ///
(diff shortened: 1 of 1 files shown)


dtolnay (approved):
Thanks!

PR #1510, opened 2023-09-14
 Add example to `Token!` documentation

ModProg:
This allows rust-analyzer to correctly complete the preferred brackets, instead of parenthesis.

https://github.com/rust-lang/rust-analyzer/issues/15608#issuecomment-1718821707

I wasn't sure what to put in there as the extensive documentation is at the module doc

### DIFF
--- a/src/token.rs
+++ b/src/token.rs
@@ -840,9 +840,25 @@ define_delimiters! {
 /// A type-macro that expands to the name of the Rust type representation of a
 /// given token.
 ///
-/// See the [token module] documentation for details and examples.
+/// It can expand to the token type and be used to construct a token from a [`Span`]:
+/// ```
+/// # use syn::{Token};
+/// use proc_macro2::Span;
+///
+/// let fn_token: Token![fn] = Token![fn](Span::call_site());
+/// ```
+/// It can be used with [`ParseBuffer::peek`]:
+/// ```
+/// # use syn::{Token, parse::ParseStream};
+/// # fn parser(input: ParseStream) {
+/// input.peek(Token![<<]);
+/// # }
+/// ```
+///
+/// See the [token module] documentation for details and more examples.
 ///
 /// [token module]: crate::token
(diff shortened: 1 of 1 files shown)


src/token.rs
```diff
@@ -840,6 +840,11 @@ define_delimiters! {
 /// A type-macro that expands to the name of the Rust type representation of a
 /// given token.
 ///
+/// ```
```
ModProg, line 843:
```suggestion
/// ```
/// # syn::Token;
```

src/token.rs
```diff
@@ -840,6 +840,12 @@ define_delimiters! {
...
+/// # syn::Token;
```
ModProg, line 844:
```suggestion
/// # use syn::Token;
```

src/token.rs
```diff
@@ -840,6 +840,12 @@ define_delimiters! {
 /// A type-macro that expands to the name of the Rust type representation of a
 /// given token.
 ///
+/// ```
+/// # use syn::Token;
+/// Token![fn]; // syn::token::Fn
+/// Token![>>]; // syn::token::Shr
```
ModProg, line 846:
```suggestion
/// # type a =
/// Token![fn]; // syn::token::Fn
/// # type b =
/// Token![>>]; // syn::token::Shr
```

dtolnay (changes requested):
Thanks for the PR!

The example you wrote would get shown to the reader like this:



which is pretty misleading as to how this macro is supposed to be used.

```console
error: macro expansion ends with an incomplete expression: expected one of `!` or `::`
   --> syn-2.0.31/src/token.rs:866:41
    |
866 |     [fn]          => { $crate::token::Fn };
    |                                         ^ expected one of `!` or `::`
    |
   ::: src/lib.rs:1:1
    |
1   | Token![fn]; // syn::token::Fn
    | ---------- in this macro invocation
```

Could you come up with a more meaningful usage to show in the example?

dtolnay (approved):

PR #1519, opened 2023-10-10
Using parse_parens, parse_braces, parse_brackets instead of macros.

rrupy:
I don't understand why we have macros parenthesized, braced and bracketed.
You cannot use '?' with them, because these macros return the ParseBuffer from expression, and in case of an error, they return it from the current function.
Simple functions are better, than macros.
```rust
let content;
if let Ok(_) = (|| Ok(bracketed!(content in input)))() { }
```
or 
```rust
if let Ok(_) = parse_braces(&input) { }
```
In first variant we have uninitialized variable.
Example from syn crate:
```rust
let content;
let (paren_token, content) = if input.peek(token::Paren) {
    (Some(parenthesized!(content in input)), &content)
} else {
    (None, input)
};
```
Perhaps we should remove these macros and make the listed functions and the values returned by them public?

### DIFF
--- a/src/attr.rs
+++ b/src/attr.rs
@@ -628,21 +628,24 @@ pub(crate) mod parsing {
     }
 
     pub(crate) fn single_parse_inner(input: ParseStream) -> Result<Attribute> {
-        let content;
+        let pound_token = input.parse()?;
+        let style = AttrStyle::Inner(input.parse()?);
+        let Brackets { token: bracket_token, content } = parse_brackets(input)?;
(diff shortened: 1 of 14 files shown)


mohe2015:
I don't know whether that would be accepted but what about just making these function public API (so remove `#[doc(hidden)]') but don't change anything else? I'm currently parsing code where I also want to return (unstable) diagnostics when parsing partially fails/succeeds and there I have a different return type so can't easily use the macros

PR #1532, opened 2023-11-07
Enable getting span locations

kromych:
Hi David,

I needed a way to extract location of the unsafe code to make use of that in the CI harness, and came up with this approach to enabling that. Here is a simple repo to validate the fork works: https://github.com/kromych/unsafe_travels.


kromych:
Found how to make this happen without re-exporting features

PR #1538, opened 2023-12-10
Add test for parsing Delimiter::None in expressions

### LINKED ISSUE #840: Synthetic test coverage of None-delimited groups
Various pieces of syntax in the parser are currently not robust to idents being wrapped by rustc in a None-delimited group, such as if the ident is associated with a macro_rules $:ident metavariable. See #838, #839. The test_round_trip.rs which is our primary source of test coverage for exotic syntax is not good at covering None-delimited groups since they never arise when parsing &amp;str input.

Similar to what we do in test_precedence.rs where we wrap every expression in the syntax tree in a parenthesized group, we could consider adding a test that wraps every ident in the input into a None-delimited group prior to parsing, and then run that over the full rust-lang/rust repo.

dtolnay:
This test will take an expression such as `(**self).get()`, replace the parens with invisible delimiters (`Delimiter::None`) i.e. equivalent to `$e.get()` where the `$e` metavariable contains `**self`, then print the syntax tree containing invisible delimiters into a TokenStream, reparse the TokenStream back to syn::Expr, and verify successful round trip.

Closes #840.

### DIFF
--- a/tests/test_precedence.rs
+++ b/tests/test_precedence.rs
@@ -128,7 +128,8 @@ fn test_expressions(path: &Path, edition: Edition, exprs: Vec<syn::Expr>) -> (us
                 continue;
             };
 
-            let syn_parenthesized_code = syn_parenthesize(expr).to_token_stream().to_string();
+            let syn_parenthesized_code =
+                syn_parenthesize(expr.clone()).to_token_stream().to_string();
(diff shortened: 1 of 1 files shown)


dtolnay:
Not a bad start:

```console
391215 passed | 1913 failed
```

Issue #1526, opened 2023-11-05
Parse `gen` blocks and functions


dtolnay:
Tracking issue: https://github.com/rust-lang/rust/issues/117078

```rust
fn moved() -> impl Iterator<Item = u32> {
    let mut x = "foo".to_string();
    gen move {
        yield 42;
        if x == "foo" { return }
        x.clear();
        for x in 3..6 { yield x }
    }
}
```

PR #1549, opened 2023-12-19
Parse gen blocks and functions

mohe2015:
https://github.com/dtolnay/syn/issues/1526


mohe2015:
Seems like some tests fail because `gen` is now a keyword. I don't know how to fix this so feel free to make changes if you want to. Otherwise this is just a partially finished PR.

dtolnay (commented):
Thanks for the PR!

This would need to be closer to stabilization before I would consider adding a syntax tree node for it. The syntax only just got introduced in rustc.

PR #1562, opened 2023-12-30
Enable rustc AST test suite not based on compiler version, but feature support

dtolnay:


dtolnay (commented):
Screenshot from 2023-12-30 11-21-56

PR #1589, opened 2024-02-14
Export TokenMarker

### LINKED ISSUE #1588: `TokenMarker` is unexported
I believe that `TokenMarker` has been accidentally been left unexported. It's `pub` while other non-exported items are `pub(crate)`. 

```rust
pub enum TokenMarker {}
```
https://github.com/dtolnay/syn/blob/master/src/lookahead.rs#L157

I ran into this when trying to make part of my code more generic, and wanted to map my types to types and items in `syn`, like this.:

```rust
impl InputVariant for MyIdent {
    const TOKEN: fn(syn::lookahead::TokenMarker) -> proc_macro2::Ident = syn::Ident
}
```

fredrik-hammar:
Fixes #1588: `TokenMarker` is unexported.

### DIFF
--- a/src/parse.rs
+++ b/src/parse.rs
@@ -201,7 +201,7 @@ use std::rc::Rc;
 use std::str::FromStr;
 
 pub use crate::error::{Error, Result};
-pub use crate::lookahead::{Lookahead1, Peek};
+pub use crate::lookahead::{Lookahead1, Peek, TokenMarker};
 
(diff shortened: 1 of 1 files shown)


dtolnay (commented):
Thanks for the PR!

I suggested an alternative in #1588. I'll close this for now, but please follow up there and we can reopen the PR as needed.

PR #1592, opened 2024-02-27
parse_parens, parse_braces, parse_brackets and structs returned by them are public API now.

rrupy:
#1519 
parenthesized!, braced! and bracketed! macros are very verbous in situations where you want to handle Err in current function beacause in this case they are expand in return statement. That's why they must be public.

Code example:

If you have an optional code in parentheses and you don't need the paren token:
```rust
if input.peek(token::Paren) {
    let paren_content;
    parenthesized!(paren_content in input);
    paren_content.parse::<ExprUnsafe>()?;
}
```
vs
```rust
if let Ok(Parens { token: _, content}) = parse_parens(input) {
    content.parse::<ExprUnsafe>()?;
}
```

### DIFF
--- a/src/group.rs
+++ b/src/group.rs
@@ -4,34 +4,33 @@ use crate::token;
 use proc_macro2::extra::DelimSpan;
 use proc_macro2::Delimiter;
 
-// Not public API.
-#[doc(hidden)]
+/// The rust code enclosed in parentheses
 pub struct Parens<'a> {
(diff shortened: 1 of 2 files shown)


dtolnay (commented):
In the code example you've shared, I think the difference is not significant enough to justify exposing these additional APIs.

Thanks anyway for the PR!

PR #1607, opened 2024-03-29
Add generator signature

QuentinPerez:
Hello 👋,

I've added support for the 'gen' keyword in the function signature. Let me know if there's anything I may have overlooked. Thank you!


dtolnay (commented):
Thanks -- but in general we do not implement eRFC syntax.

PR #1608, opened 2024-03-30
Use `dep:` and optional dependency feature `?` syntax

BD103:
This PR comes in 2 parts:

The first uses the `dep:` syntax for the `printing` feature depending on `quote`, which removes the unused implicit feature.

The second uses the `?` syntax in the `proc-macro` feature, so `quote` is not pulled in as a dependency if `printing` is not enabled.

Unfortunately, both of these changes raise the MSRV back to 1.60. Before finishing my changes in CI, I wanted to confirm that this was fine.


dtolnay (commented):
1.60 is fine. Thanks!

dtolnay (approved):

BD103:
Thank you! (That was fast 😄)

dtolnay:
Published in 2.0.57.

joshlf:
Just a heads up that the MSRV bump broke zerocopy: https://github.com/google/zerocopy/issues/1085. I know that there's no ecosystem consensus on whether MSRV bumps should be considered breaking changes; just wanted to provide a data point.

If syn's MSRV policy is to not consider MSRV bumps to be breaking changes, it'd be great if that were documented somewhere so we'd know to be aware of the risk.

PR #1683, opened 2024-06-20
List peekable token types in documentation of Peek trait

dtolnay:

### DIFF
--- a/src/ext.rs
+++ b/src/ext.rs
@@ -4,7 +4,6 @@ use crate::buffer::Cursor;
 use crate::error::Result;
 use crate::parse::ParseStream;
 use crate::parse::Peek;
-use crate::sealed::lookahead;
 use crate::token::CustomToken;
 use proc_macro2::Ident;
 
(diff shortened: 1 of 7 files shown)


fubupc:
May I ask a design question here?

I’ve been looking at the design for `Peek` trait and `ParseStream::peek` method and I couldn’t find clear guidance in the docs. The `token` parameter doesn't seem to be used at all and the only its type appears to matter.  Then why not simply remove the `token` parameter from `ParseStream::peek` as follows:

```Rust
impl ParseStream {
    pub fn peek<T: Token>(&self) -> bool {
        T::peek(self.cursor())
    }
}
```

Is the parameter kept for ergonomic reasons, to avoid the need of explict type annotations (e.g., having to write `ParseStream::peek::<Lit>()`) ? Is the purpose of the `Peek` trait and the functions named after token types (e.g., `fn Lit(..)`) simply to simulate passing a type (rather than a value) as an argument to `peek`, given that Rust doesn’t natively support using types as function parameters?

PR #1706, opened 2024-07-12
Remove mut from Puntuated::get()

lemunozm:
Just removing the mutable requirement for `Puntuacted::get()`, probably a typo.


dtolnay (approved):
Thanks!

Issue #1677, opened 2024-06-19
Parse precise capturing syntax


dtolnay:
RFC: https://github.com/rust-lang/rfcs/pull/3617
Tracking issue: https://github.com/rust-lang/rust/issues/123432

```rust
fn f<'a, T>() -> impl Sized + use<'a, T> {}
```

traviscross:
Just an update here that the [stabilization report](https://github.com/rust-lang/rust/pull/127672#issue-2406639114) for this has gone up.

The earliest this syntax could conceivably stabilize is with Rust 1.82, which will branch on 2024-08-30 and will be released on 2024-10-17.

PR #1707, opened 2024-07-13
Parse `use<'a, T>` precise capturing bounds as verbatim

compiler-errors:
Parses the new `use<>` precise capturing bound (https://rust-lang.github.io/rfcs/3617-precise-capturing.html) as verbatim in all bound positions.

Not exactly sure what the process is for syn support for nightly features. I could alternatively prepare a PR adding full support for this.

cc https://github.com/rust-lang/rust/issues/123432

### DIFF
--- a/src/generics.rs
+++ b/src/generics.rs
@@ -772,6 +772,27 @@ pub(crate) mod parsing {
 
             let begin = input.fork();
 
+            if cfg!(feature = "full") && input.peek(Token![use]) {
+                input.parse::<Token![use]>()?;
+                input.parse::<Token![<]>()?;
+                while !input.peek(Token![>]) {
+                    if input.peek(Lifetime) {
+                        input.parse::<Lifetime>()?;
+                    } else if input.peek(Ident) {
+                        input.parse::<Ident>()?;
+                    } else {
+                        break;
+                    }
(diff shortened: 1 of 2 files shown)


src/generics.rs
```diff
@@ -772,6 +772,27 @@ pub(crate) mod parsing {
 
             let begin = input.fork();
 
+            if cfg!(feature = "full") && input.peek(Token![use]) {
```
compiler-errors, line 775: I've gated this behind `cfg(feature = "full")` since that's what `~const` does below.

src/generics.rs
```diff
@@ -772,6 +772,27 @@ pub(crate) mod parsing {
...
+                input.parse::<Token![use]>()?;
+                input.parse::<Token![<]>()?;
+                while !input.peek(Token![>]) {
```
compiler-errors, line 778: Opted for a manual loop here since I'm not certain `Punctuated` gives me the right API for this.

traviscross:
This fixes https://github.com/dtolnay/syn/issues/1677.   The stabilization report for this is up [here](https://github.com/rust-lang/rust/pull/127672#issue-2406639114).

cc @dtolnay

dtolnay (approved):
Thanks!

PR #1736, opened 2024-09-23
feat: make parse_str and parse_file accept str-like

fowles:


dtolnay (commented):
Thanks for the PR, but I prefer the existing API.

PR #1854, opened 2025-03-30
feat: ParseBuffer::call takes FnOnce(..) instead of fn(..)

aatifsyed:
Curious how you'd feel about this.
Reasonably sure it's backwards compatible, and would be great for users of [`attrs`](https://docs.rs/attrs).


aatifsyed:
Ah wait I'm silly, this isn't necessary at all

PR #1859, opened 2025-04-16
Added syn::Generics::is_empty

JohnScience:
This method is useful to check whether some item is generic.


dtolnay (commented):
Thanks! Would you be able to share how you plan to use this?

PR #1861, opened 2025-04-26
Fix lifetime on `as_turbofish`

JakobDegen:
I'd like to write something like this:

```rust
struct W(syn::Generics);

impl W {
    pub fn as_turbofish(&self) -> syn::Turbofish<'_> {
        self.0.split_for_impl().1.as_turbofish()
    }
}
```

Unfortunately, that doesn't compile today and as far as I can tell it's completely unimplementable -
the lifetime of the `syn::Turbofish` ends up tied to the temporary returned by `split_for_impl`.

That seems silly and unintentional, fix it.


dtolnay (approved):
Thanks!

PR #1862, opened 2025-05-04
Include semicolon in doc comment example for `syn::Local`.

martician1:
Include a semicolon in `syn::Local`’s doc comment example to better reflect the struct fields.


dtolnay (approved):
Thanks!

PR #1851, opened 2025-03-06
Support parsing `#![feature(default_field_values)]`

estebank:
- RFC: https://github.com/rust-lang/rfcs/pull/3681
- Tracking issue: https://github.com/rust-lang/rust/issues/132162
- Feature gate: `#![feature(default_field_values)]`

```rust
#[derive(Default)]
struct Pet {
    name: Option<String>,
    age: i128 = 42,
    //        ^^^^
}
```

Fix #1774.

Issue #1774, opened 2024-10-29
Parse default values in fields


dtolnay:
RFC: https://github.com/rust-lang/rfcs/pull/3681

```rust
#[derive(Default)]
struct Pet {
    name: Option<String>,
    age: i128 = 42,
    //        ^^^^
}
```

PR #1870, opened 2025-06-12
Parse default field values as `Type::Verbatim`

DJMcNab:
See #1774 (noting that a true fix will require 3.x and #1851)

* RFC: https://github.com/rust-lang/rfcs/pull/3681
* Tracking issue: https://github.com/rust-lang/rust/issues/132162
* Feature gate: `#![feature(default_field_values)]`

```rust
#[derive(Default)]
struct Pet {
    name: Option<String>,
    age: i128 = 42,
    //        ^^^^
}
```

Adapted from #1851. cc @estebank

This PR is an alternative to #1851 but avoiding the breaking-change. Instead, if a named field has a default value, the field's entire type and the default value is parsed as a `Type::Verbatim`. Our users can then detect and support this case, by re-parsing the verbatim type into the type and default value expression.

Semantically of course, the default value is not part of the type, but syntactically, the default value is in type position. Our handling of negative inherent impls has a similar shape.

The interaction with `unnamed_fields` was a bit unclear to me, so I took my cue from #1851 (of allowing both features to be combined). Note that since https://github.com/rust-lang/rust/pull/131045, `unnamed_fields` is not supported by upstream Rust.

I chose not to implement the solution you proposed in https://github.com/dtolnay/syn/pull/1851#discussion_r1983947505, both because the idiomatic implementation strategy was unclear to me, and because if I understand correctly, the proposed solution makes supporting `default_field_values` in dependent macros unfeasible.

Thanks for all your great work!

### DIFF
--- a/src/data.rs
+++ b/src/data.rs
@@ -334,6 +334,7 @@ pub(crate) mod parsing {
 
             let colon_token: Token![:] = input.parse()?;
 
+            let type_start = input.fork();
             let ty: Type = if unnamed_field
                 && (input.peek(Token![struct])
                     || input.peek(Token![union]) && input.peek2(token::Brace))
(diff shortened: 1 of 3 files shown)


jplatte:
Gentle ping, @dtolnay

PR #1880, opened 2025-08-10
Reorder TraitBound fields to match rustc AST

npmccallum:
I have fixed the upstream bug which led to this mistake:

https://github.com/rust-lang/rust-analyzer/pull/20417

### DIFF
--- a/src/generics.rs
+++ b/src/generics.rs
@@ -412,9 +412,9 @@ ast_struct! {
     #[cfg_attr(docsrs, doc(cfg(any(feature = "full", feature = "derive"))))]
     pub struct TraitBound {
         pub paren_token: Option<token::Paren>,
-        pub modifier: TraitBoundModifier,
         /// The `for<'a>` in `for<'a> Foo<&'a T>`
         pub lifetimes: Option<BoundLifetimes>,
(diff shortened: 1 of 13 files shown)


src/generics.rs
```diff
@@ -845,8 +845,8 @@ pub(crate) mod parsing {
     #[cfg_attr(docsrs, doc(cfg(feature = "parsing")))]
     impl Parse for TraitBound {
         fn parse(input: ParseStream) -> Result<Self> {
-            let modifier: TraitBoundModifier = input.parse()?;
             let lifetimes: Option<BoundLifetimes> = input.parse()?;
+            let modifier: TraitBoundModifier = input.parse()?;
```
dtolnay, line 849: I don't think this change is correct. As far as I know, there have only ever been 2 kinds of rustc: one where `?for<'a> Trait` is valid trait bound syntax (all versions prior to 1.82) and one where neither `?for<'a> Trait` nor `for<'a> ?Trait` is valid trait bound syntax. There has never been any version of Rust where the syntax implemented here is valid trait bound syntax.

Could you provide some more justification about this change?

dtolnay (changes requested):

Issue #1911, opened 2025-08-16
Add `default` field for default value in `Field`


dtolnay:
For the language feature tracked by this issue:

- https://github.com/rust-lang/rust/issues/132162

```rust
#[derive(Default)]
struct Pet {
    name: Option<String>,
    age: i128 = 42,
}             ^^^^
```

Issue #1889, opened 2025-08-13
Parse `super let` syntax


dtolnay:
- Tracking issue: https://github.com/rust-lang/rust/issues/139076

```rust
let a = {
    super let b = temp();
    &b
};
```

PR #1942, opened 2025-11-07
Parse super let syntax (#1889)

AMDmi3:
I'm interested in some unstable rust features (in particular, this one and #1911), but these are frequently blocked by absence of support in `syn` (via other crates, such as `async-trait` or `serde`), so I've decided to give it a try in adding the required support. I haven't dived deep into the code yet, so I'm still not sure about some parts:
- Condition in `parse_stmt`, `if input.peek(Token![let]) && !input.peek(token::Group)` - does it make sense, can both these peeks return true? Is it correct after my change? What's `token::Group` anyway?
- `ParseStream::parse()` return value seems to be coercible to `Option<>`, can part in `stml_local` be rewritten as simply `let super_token: Option<Token![super]> = input.parse()?;`?
- Most importantly, is this correct approach overall, or do we need maybe a separate `Stmt` kind for `super let`? Is my approach API safe, as presumably client code which processes Local nodes may be unaware of `super_token` addition? Is either approach future safe if, say, `super let` syntax changes?
- Should `Local` docstring be updated to mention `super let`?

### DIFF
--- a/src/gen/clone.rs
+++ b/src/gen/clone.rs
@@ -1389,6 +1389,7 @@ impl Clone for crate::Local {
     fn clone(&self) -> Self {
         crate::Local {
             attrs: self.attrs.clone(),
+            super_token: self.super_token.clone(),
             let_token: self.let_token.clone(),
             pat: self.pat.clone(),
             init: self.init.clone(),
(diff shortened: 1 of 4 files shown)


dtolnay (commented):
Thank you, but this would need to be closer to stabilization as a language feature before I would consider building it into the syntax tree.

PR #1943, opened 2025-11-08
Reduce repetitions of path.segments.first/last.ident.span()

dishmaker:
This reduces llvm-lines only by 126  on `rustc 1.91.0`


dtolnay (commented):
I think the original code is fine to keep as-is.

PR #1955, opened 2026-01-05
Support Default Field Values, backwards compatibly

### LINKED ISSUE #1774: Parse default values in fields

DJMcNab:
This PR incorporates work from @estebank in #1851. See also https://github.com/rust-lang/rust/issues/132162#issuecomment-3647488349 for context; this is the PR for "solution 2", to unblock the language feature.

Fixes #1774

* RFC: https://github.com/rust-lang/rfcs/pull/3681
* Tracking issue: https://github.com/rust-lang/rust/issues/132162
* Feature gate: `#![feature(default_field_values)]`

```rust
#[derive(Default)]
struct Pet {
    name: Option<String>,
    age: i128 = 42,
    //        ^^^^
}
```

By running `cargo-semver-checks --all-features` (v0.45.0), I have confirmed that this is not a breaking change.

This PR has two related changes:
1) It exposes `FieldWithDefault`, which is a version of `Field`, with the addition of the optional `default` field. This is then used to create a parallel hierarchy of all Syn types, with the same naming pattern. That is, it also creates `FieldsNamedWithDefault`, `FieldsWithDefault`, `VariantsWithDefault`, `DataStructWithDefault`, etc.
2) The parsing for this is shared with `Field`, to allow for as much shared code as possible. There is one subtle change here; the parsing for `Field` (and therefore `FieldsNamed`, `Fields`, `DeriveInput`, etc.) will now ignore any default value provided (i.e. not give an error, but also not reflect it in the AST). This enables a silent upgrade of most derive macros to support their same codegen as if the default field values didn't exist, which will be extremely valuable in the ecosystem.

This is explicitly not https://github.com/dtolnay/syn/issues/1911; instead, that issue would likely be updated to mean "revert this PR, and land #1851 as part of v3.x".

The new structs in this PR is *not* behind a feature flag. This is because the current infrastructure in `syn-internal-codegen` only supports "`any`" style feature flags, whereas this would need to be handled as "all" style feature flags.
I'm not going to make the changes to the codegen infrastructure myself. I would move these types behind feature flags if:
- You say that you want this code to be in its own modules (and therefore under two feature flags); OR
- You provide the patches to the infrastructure which means it supports the "all" style flags.

The silent ignoring of the default values would definitely be reasonable to object to. I think it's the least bad option on an ecosystem level, but I'm happy to change it (either some form of logging, or just throwing an error).

The parallel hierarchy is not complete. That is, we do not support structs with default field values inside blocks (i.e. inside functions/block expressions/etc.). That's because it requires changing so much more code. (But obviously if they exist, they'll be silently ignored).

A reasonable alternative would be to implement #1870 for those nested fields.

### DIFF
--- a/examples/heapsize/heapsize_derive/src/lib.rs
+++ b/examples/heapsize/heapsize_derive/src/lib.rs
@@ -2,13 +2,14 @@ use proc_macro2::TokenStream;
 use quote::{quote, quote_spanned};
 use syn::spanned::Spanned;
 use syn::{
-    parse_macro_input, parse_quote, Data, DeriveInput, Fields, GenericParam, Generics, Index,
+    parse_macro_input, parse_quote, DataWithDefault, DeriveInputWithDefault, FieldsWithDefault,
+    GenericParam, Generics, Index,
 };
 
 #[proc_macro_derive(HeapSize)]
 pub fn derive_heap_size(input: proc_macro::TokenStream) -> proc_macro::TokenStream {
     // Parse the input tokens into a syntax tree.
-    let input = parse_macro_input!(input as DeriveInput);
+    let input = parse_macro_input!(input as DeriveInputWithDefault);
 
     // Used in the quasi-quotation below as `#name`.
     let name = input.ident;
@@ -44,11 +45,11 @@ fn add_trait_bounds(mut generics: Generics) -> Generics {
 }
 
 // Generate an expression to sum up the heap size of each field.
-fn heap_size_sum(data: &Data) -> TokenStream {
+fn heap_size_sum(data: &DataWithDefault) -> TokenStream {
     match *data {
-        Data::Struct(ref data) => {
+        DataWithDefault::Struct(ref data) => {
             match data.fields {
-                Fields::Named(ref fields) => {
+                FieldsWithDefault::Named(ref fields) => {
                     // Expands to an expression like
                     //
                     //     0 + self.x.heap_size() + self.y.heap_size() + self.z.heap_size()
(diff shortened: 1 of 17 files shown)


estebank:
Gentle ping @dtolnay, in case you can quickly comment on the above open question even if you don't have the time to perform a full review.

dtolnay (commented):
> the parsing for `Field` (and therefore `FieldsNamed`, `Fields`, `DeriveInput`, etc.) will now ignore any default value provided (i.e. not give an error, but also not reflect it in the AST).

I feel that this is misconceived. This is going to result in attributes silently erasing default values written in the source code. For example the following program would print `1 0` instead of `1 1`. Discovering occurrences of this in a large codebase is going to be difficult, and making sense of why their default value is not working is going to be difficult for users. I would expect instead an error explaining that code is being parsed by a macro that does not support default field values, with guidance what change to make to the macro.

```rust
#[derive(Default)]
struct Outer {
    x: u8 = 1,
}

#[tracing::instrument]
fn main() {
    #[derive(Default)]
    struct Inner {
        x: u8 = 1,
    }

    println!("{}", Outer::default().x);
    println!("{}", Inner::default().x);
}
```

Techcable:
I don't know if this has been suggested before, but maybe add a `FieldWithDefault` variant to the `Type` enum?

This will break macros that quote the field type, but it's technically not a breaking change.

PR #1956, opened 2026-01-06
Disable std prelude

tamird:
This makes visible the dependency on std prelude macros such as
`error::Error`. Helpful for planning how to do no-std.

Tested with `cargo hack clippy --feature-powerset -- --deny warnings`
(all 2112 combinations).

See also https://github.com/dtolnay/proc-macro2/pull/528.


dtolnay (approved):
Thanks!

PR #1960, opened 2026-01-20
expose "safe fn" in extern block as ForeignItemFn

Wasabi375:
Currently when parsing a "safe fn" `syn` will represent this as a `ForeignItem::Verbatim`.
This makes it hard to use them in proc-macros. 

Right now `parse_signature` will return `Ok(Some(sig))` for signatures without the "safe" keyword. If the "safe" keyword is found it returns `Ok(None)` instead. 
If `None` is returned `ForeignItem::parse` will assume that the function signature contains the "safe" keyword and returns a `Verbatim`.

I see 2 possible solutions to improve this:
1. add `Option<Safe>` to `Signature`
2. change `parse_signature` to return `Ok(sig, Some(safe))` and store the "safe" keyword in the `ForeignItemFn`
3. Introduce a `ForeignSignature` that extends `Signature` with the optional safe keyword

There are some Pros/Cons to all 3 approaches
1. The "safe" keyword is only valid within an `extern` block so I am not sure if this is a good abstraction. However this would be the simplest to implement.
2. Still simple to implement, but it stores the "safe" keyword as a field on `ForeignItemFn` which means that there needs to be special handling in `ForeignItemFn::to_tokens` to ensure that the "safe"  keyword is inserted at the correct position.
3. This is the cleanest approach, as it properly represents valid rust code. However it is probably the solution with the most code. In any case I decided to implement this. Option 1 and 2 feel like hacks to me.

Open questions:
1. Are there any tests I should add for this. If so can you point me to an existing test that I could use to orient myself. I'm not sure I fully understand the test framework here.
2. I'm not sure how you handle versioning. This change changes the fields in `ForeignItemFn` and I guess that is a breaking change. However I can't imagine you release a breaking version change every time rust adds a new feature with new syntax. No idea how you want to handle something like this.

### DIFF
--- a/src/gen/clone.rs
+++ b/src/gen/clone.rs
@@ -879,6 +879,19 @@ impl Clone for crate::FnArg {
 }
 #[cfg(feature = "full")]
 #[cfg_attr(docsrs, doc(cfg(feature = "clone-impls")))]
+impl Clone for crate::ForeignFnSafety {
+    fn clone(&self) -> Self {
+        match self {
+            crate::ForeignFnSafety::Unsafe(v0) => {
+                crate::ForeignFnSafety::Unsafe(v0.clone())
+            }
+            crate::ForeignFnSafety::Safe(v0) => crate::ForeignFnSafety::Safe(v0.clone()),
+            crate::ForeignFnSafety::None => crate::ForeignFnSafety::None,
+        }
+    }
+}
+#[cfg(feature = "full")]
+#[cfg_attr(docsrs, doc(cfg(feature = "clone-impls")))]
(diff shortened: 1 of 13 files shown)


nik-rev:
> I did not know that and the rust reference specifically states that this is not actually allowed. 

I got interested in this and looked it up.

https://doc.rust-lang.org/reference/items/functions.html#footnote-extern-safe

Semantically disallowed, syntactically allowed. Macros only operate on the syntactic level:

> The safe function qualifier is only allowed semantically within extern blocks. [↩](https://doc.rust-lang.org/reference/items/functions.html#fr-extern-safe-1)

PR #1974, opened 2026-03-22
Feature: add support for `impl const Trait` syntax

usmanovbf:
This PR introduces support for parsing constant trait implementations, specifically the `const impl` and `impl const` syntaxes. This resolves the parsing error reported in issue #1972 and aligns `syn` with the experimental syntax introduced by RFC 2632 (Const Trait Impls).

**Changes included in this PR:**
* **AST Modification:** Added a new `constness: Option<Token![const]>` field to the `syn::ItemImpl` struct.
* **Parsing Logic (`Parse`):** Updated the parser in `src/item.rs` to correctly recognize and consume the `const` keyword. It handles both `const impl Trait` and `impl const Trait` variations to ensure robust compatibility with the experimental compiler features.
* **Token Serialization (`ToTokens`):** Updated the implementation to assure the `const` token is properly preserved and written back into the token stream during code generation.
* **Auto-generated AST Traversal:** Updated all generated visitor and folder modules (`clone.rs`, `debug.rs`, `eq.rs`, `fold.rs`, `hash.rs`, `visit.rs`, `visit_mut.rs`) to account for the new `constness` field.
* **Testing:** Added comprehensive test coverage in `tests/test_item.rs` to verify successful parsing, token round-tripping, and correct behavior when combined with other modifiers like `default` and `unsafe`.

**SemVer Note:** 
Please note that adding a public field to `ItemImpl` is a breaking change for users relying on exhaustive pattern matching. I defer to the maintainers on the appropriate milestone or breaking release batch for this PR.

### DIFF
--- a/src/gen/clone.rs
+++ b/src/gen/clone.rs
@@ -1187,6 +1187,7 @@ impl Clone for crate::ItemImpl {
             attrs: self.attrs.clone(),
             defaultness: self.defaultness.clone(),
             unsafety: self.unsafety.clone(),
+            constness: self.constness.clone(),
             impl_token: self.impl_token.clone(),
             generics: self.generics.clone(),
             trait_: self.trait_.clone(),
--- a/src/gen/debug.rs
+++ b/src/gen/debug.rs
@@ -1751,6 +1751,7 @@ impl crate::ItemImpl {
         formatter.field("attrs", &self.attrs);
         formatter.field("defaultness", &self.defaultness);
         formatter.field("unsafety", &self.unsafety);
+        formatter.field("constness", &self.constness);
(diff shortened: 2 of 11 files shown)


dtolnay (commented):
Thank you, but this would need to be closer to stabilization as a language feature before I would consider building it into the syntax tree.

orium:
This is currently used in the stdlib (https://doc.rust-lang.org/stable/src/alloc/vec/mod.rs.html#918), so currently syn cannot be used to parse the stdlib. Just to make the decision to not support it takes that into consideration.

dtolnay:
Parsing unstable standard library code is not a goal of this crate.

PR #2063, opened 2026-07-19
Parse const impl

### LINKED ISSUE #1980: Parse const impl syntax
Since https://github.com/rust-lang/rust/pull/158009 the syntax for implementing const trait changed from `impl const Trait for Thing` to `const impl Trait for Thing`.

Example: https://github.com/rust-lang/rust/blob/d39561c2d2b55985d3b6331cc52403e038e7fc8b/compiler/rustc_data_structures/src/unord.rs#L260

```rust
const impl<V: Eq + Hash> Default for UnordSet<V> {
```

dtolnay:
Closes https://github.com/dtolnay/syn/issues/1980.


nazar-pc (commented):
It is quite inconvenient that regular and const impls are represented by separate types. This is especially true because const traits are not required to be implemented in const way, meaning code that can support either will have to create wrapper enums to handle both in the same way.

PR #2080, opened 2026-09-02
In LitStr::parse_with, report correct span for lex errors

sunshowers:
In proc macros, we were previously falling back to `Span::call_site` in this case. The other errors in this function return correct spans, so match that here.

(Not sure how to write a test for this in this repo, but I patched it into a dependency and it seems to work.)


dtolnay (approved):
Thanks!

dtolnay:
Published in 3.0.5.

